PR #2, opened 2021-08-03
Explicit arr object creation

Nlte:
Hi, I have started playing with the cookbook examples and there is a little doc section that I did not find clear hence here this PR.

### Issue
The **testsetup::** blocs in the rst files are not written out in the resulting HTML.
In the "Given an array with 100 numbers, from 0 to 99" sections the **arr** object pops up ex nihilo in the documentation which might lose the reader a bit even though most people will guess the np.arange here.

### Proposed solution
Add the array creation code in the testcode:: blocs for those sections section to make it more explicit.

That is going from
```
Given an array with 100 numbers, from 0 to 99

print(f"{arr[0]} .. {arr[-1]}")

0 .. 99
```

to

```
Given an array with 100 numbers, from 0 to 99

import numpy as np
import pyarrow as pa

arr = pa.array(np.arange(100))

print(f"{arr[0]} .. {arr[-1]}")

0 .. 99
```
in the resulting HTML.

Thanks,
Nathanaël

### DIFF
--- a/python/source/io.rst
+++ b/python/source/io.rst
@@ -21,6 +21,11 @@ Given an array with 100 numbers, from 0 to 99
 
 .. testcode::
 
+    import numpy as np
+    import pyarrow as pa
+
+    arr = pa.array(np.arange(100))
+
     print(f"{arr[0]} .. {arr[-1]}")
 
 .. testoutput::
@@ -121,6 +126,11 @@ the Arrow IPC format.
 Given an array with 100 numbers, from 0 to 99
 
 .. testcode::
+
+    import numpy as np
+    import pyarrow as pa
+
+    arr = pa.array(np.arange(100))
 
     print(f"{arr[0]} .. {arr[-1]}")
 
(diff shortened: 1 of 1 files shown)


westonpace:
@amol- mind taking a look?

thisisnic:
FYI, I think amol- is on vacation for another week or so, so this won't get looked at til then

jorisvandenbossche:
I had the same observation while going through the IO cookbook: I think it would indeed be nice that the example is fully reproducible (copy/pastable), and thus includes the code to create the dummy data.

The same code in the `testsetup` could then be removed?

Nlte:
I think we still need the `testsetup` code for the intermediate cell tests to pass.

westonpace:
@amol- any thoughts on this?  Otherwise I'll go ahead and merge this evening.

westonpace (approved):
Looks good.  I agree with Joris.  For languages like python I think there is some expectation that an example be "runnable" so I like this even if it is a bit more verbose.

westonpace:
Merging.  @Nlte Thank you very much for the contribution!  Feel free to add any other ideas.  This project is just getting started so the more input the better :)

PR #3, opened 2021-08-04
Refactor GH Actions job that uses third party action

thisisnic:
This PR refactors the checkout-and-deploy job so it no longer uses a third party GH Action (which are not allowed on Apache repos unless specifically authorised)

### DIFF
--- a/.github/workflows/deploy_cookbooks.yml
+++ b/.github/workflows/deploy_cookbooks.yml
@@ -11,38 +11,39 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v1
-      - uses: r-lib/actions/setup-r@v1
-      - uses: r-lib/actions/setup-pandoc@v1
+      - name: Setup R
+        uses: r-lib/actions/setup-r@v1
+      - name: Setup pandoc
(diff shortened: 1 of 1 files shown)


westonpace:
This looks good, I was able to get it to build on my fork.  I'll merge.  I wonder if we might want to do something slightly different though.

This will checkout the existing gh-pages branch, update it, and then push.  However, this has a few downsides:

* It will fail if the gh-pages branch doesn't exist
* It will keep a git history for gh-pages (I don't think we want this for built artifacts)
* It might leave behind files that get removed (I think the current "add all" will only capture new or modified files)

Alternatively, we could checkout the main branch initially (just to get origin setup), use "checkout --orphan" to create a brand new slate.  Do "add all / commit".  Force push to gh-pages.

If this sounds good to you I'd be happy to work it through.

thisisnic:
> Alternatively, we could checkout the main branch initially (just to get origin setup), use "checkout --orphan" to create a brand new slate. Do "add all / commit". Force push to gh-pages.
>
> If this sounds good to you I'd be happy to work it through.

Yeah, your approach definitely sounds like an improvement, that'd be great! :)

PR #7, opened 2021-08-05
remove duplicated sections and add ASF license in .asf.yaml

thisisnic:
This PR removes the second "github" section header from the config file and adds the Apache license


westonpace (approved):
LGTM

PR #18, opened 2021-08-06
#17 no pacman

### LINKED ISSUE #17: R build failing with Error: `Error in loadNamespace(x) : there is no package called ‘testrmd’`
I was hoping this was just intermittent since there was no actual code change but it's happened 3 times in a row now so I'm not sure what could have caused it.

thisisnic:
Fixes #17 and removes 2 unnecessary dependencies

### DIFF
--- a/r/content/index.Rmd
+++ b/r/content/index.Rmd
@@ -5,7 +5,6 @@ params:
 ---
 
 ```{r setup, include = FALSE}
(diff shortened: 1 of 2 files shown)


westonpace:
I don't understand the R but it's probably not a big deal and it will hopefully fix the build so I'll go ahead and merge.

PR #20, opened 2021-08-06
#13 install latest release

### LINKED ISSUE #13: Change builds to use prebuilt binaries
Right now the CI build is pretty slow and a fair amount of time is spent downloading / installing build dependencies and building Arrow.  We should be able to use prebuilt binaries for this.

This is somewhat related to #10 .  Do we grab the last nightly build or the last released build?

thisisnic:
Fixes #13   - When installing the R dependencies, the default behaviour is now to install the latest released version of the Arrow R package.

### DIFF
--- a/r/scripts/install_dependencies.R
+++ b/r/scripts/install_dependencies.R
@@ -49,7 +49,7 @@ get_installed_version <- function(pkg) {
 #' @param pkg Package name, character.
 load_package <- function(pkg) {
   if (!suppressWarnings(suppressMessages(require(pkg, character.only = TRUE)))) {
-    install.packages(pkg, repos = "https://cran.rstudio.com")
+    install.packages(pkg)
   }
   library(pkg, character.only = TRUE)
 }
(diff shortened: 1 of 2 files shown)


r/scripts/install_dependencies.R
```diff
@@ -1,41 +1,85 @@
...
-    pkg_url <- paste0("https://cran.r-project.org/src/contrib/Archive/arrow/arrow_", build_version, ".tar.gz")
-    install.packages(pkg_url, repos = NULL, type = "source")
+
+  # Only install the latest released version if it's not already installed
+  if (version_to_install == latest_release && installed_version != latest_release) {
+    Sys.setenv(NOT_CRAN = TRUE)
+    install.packages("arrow")
+    # Otherwise install the build version specified if not already installed
+    # TODO: refactor this to install the specific version from the nightlies if
+    # a binary is available
```
westonpace, line 73: Nit, can you add a GH issue for this and reference it as part of the todo, `TODO(#17)` or something like that.

westonpace (approved):
One small nit and then I'll merge.  If we get more R interest later they can always make changes later.

r/scripts/install_dependencies.R
thisisnic: Good point!  Have updated now.

PR #22, opened 2021-08-07
Initial C++ cookbook

westonpace:
There are probably quite a few things to point out for discussion and jot down for future issues:

 * The C++ cookbook CI build is separate from the R & Python builds right now.  I think that is probably a good thing.  It is fairly involved and installs conda and other dependencies the other builds probably don't care about.
 * The C++ cookbook is not part of make all since it requires conda which may not be present on user systems.  I think in the future make all could be updated to detect conda and, if present, build C++ or skip if not present (Apache Thrift does something like this)
 * The C++ cookbook uses a custom Sphinx extension which is a bit of maintenance but, on the other hand, I couldn't find a really good plugin for rendering AND testing C++ code blocks with Sphinx.
 * The build uses "last released Arrow" and should use nightly (or multiple Arrow versions)
 * I think eventually BeginRecipe and EndRecipe will become macros so they can include "minimum supported version".  I'd really like them to be able to define their own functions so you can do something like...
```
BEGIN_RECIPE(ReturnNotOk, 3.0.0)
void HelperFunc() {

}

void Recipe() {

}
END_RECIPE(ReturnNotOk)
```
...but also do something like this (no helper functions)...
```
BEGIN_RECIPE(Builder, 3.0.0)
arrow::IntBuilder builder;
END_RECIPE(Builder)
```
* Placeholder for other things I'm sure I've forgotten


westonpace:
Example: https://westonpace.github.io/arrow-cookbook/cpp/index.html

cpp/CONTRIBUTING.md
```diff
@@ -0,0 +1,184 @@
...
+Each snippet source file contains a set of
+[googletest](https://github.com/google/googletest) tests.  Feel free to
+use any googletest features needed to help setup and verify your test.
+To reference a snippet you need to surround it in `BeginRecipe` and
+`EndRecipe` calls.  For example:
+
+```
+StartRecipe("CreatingArrays");
+arrow::Int32Builder builder;
+ASSERT_OK(builder.Append(1));
```
lidavidm, line 68: We should note that macros like ASSERT_OK come from Arrow (and ideally link to documentation for them).

build/index.html
```diff
@@ -40,6 +40,7 @@ <h1>Apache Arrow Cookbook</h1>
 		<ul>
 			<li><a href="py/index.html">Python Cookbook</a></li>
 			<li><a href="r/index.html">R Cookbook</a></li>
+			<li><a href="cpp/index.html">C++ Cookbook</a></li>
```
lidavidm, line 43: nit: why not alphabetically order?

cpp/CONTRIBUTING.md
```diff
@@ -0,0 +1,184 @@
...
+they suit that section or you can create new sections by adding
+additional ``.rst`` files in the ``source`` directory. You just
+need to remember to add them to the ``index.rst`` file in the
+``toctree`` for them to become visible.
+
+Referencing a C++ Snippet
+-------------------------
+
+Most recipes will reference a snippet of C++ code.  For simplicity
+a custom `recipe` directive htat can be used like so:
```
lidavidm, line 35:
```suggestion
a custom `recipe` directive that can be used like so:
```

cpp/CONTRIBUTING.md
```diff
@@ -0,0 +1,184 @@
...
+
+The examples should be as simple as possible.  If complex code (e.g. templates) can be
+used to do something more efficiently then there should be a simple, inefficient version
+alongside the more complex version.
+
+Do not use `auto` in any of the templates unless you must (e.g. lambdas).  Cookbook
+viewers will be using a browser, not an IDE, and it is not always simple to determine
+the inferred type.
+
+For conciseness, `using namespace std;` should be used.
```
lidavidm, line 161: I'm not sure the conciseness buys us much compared with potential name clashes/confusion over where symbols come from. Also, I don't see this rule in the Arrow source and in fact the only instance of this declaration is in vendored third-party code.

lidavidm: It also explicitly contradicts the C++ style guide: https://google.github.io/styleguide/cppguide.html#Namespaces

cpp/code/CMakeLists.txt
```diff
@@ -0,0 +1,47 @@
...
+FetchContent_MakeAvailable(googletest)
+
+# Add Arrow
+find_package(Arrow REQUIRED)
+
+# Create test targets
+enable_testing()
+include(GoogleTest)
+add_executable(
+        creating_arrow_objects
```
lidavidm, line 24: Do we want to set up cmake-format (eventually)? At least, there's inconsistent indentation in this file.

cpp/code/main.cc
```diff
@@ -0,0 +1,42 @@
...
+
+int main(int argc, char** argv) {
+  testing::InitGoogleTest(&argc, argv);
+  int retval = RUN_ALL_TESTS();
+  if (retval == 0 && HasRecipeOutput()) {
+    arrow::Status st = DumpRecipeOutput("recipes_out.arrow");
+    if (!st.ok()) {
+      std::cerr << "Tests ran successfully but failed to dump recipe output: " << st
+                << std::endl;
+      // return -1;
```
lidavidm, line 35: Was this meant to be commented out?

cpp/source/conf.py
```diff
@@ -0,0 +1,72 @@
...
+# This pattern also affects html_static_path and html_extra_path.
+exclude_patterns = []
+
+
+# -- Options for HTML output -------------------------------------------------
+
+# The theme to use for HTML and HTML Help pages.  See the documentation for
+# a list of builtin themes.
+#
+html_theme = 'alabaster'
```
lidavidm, line 64: Should we (eventually) use the same style as the Arrow docs?

cpp/code/creating_arrow_objects.cc
```diff
@@ -0,0 +1,57 @@
...
+
+using namespace std;
+
+TEST(CreatingArrowObjects, CreateArrays) {
+  StartRecipe("CreatingArrays");
+  arrow::Int32Builder builder;
+  ASSERT_OK(builder.Append(1));
+  ASSERT_OK(builder.Append(2));
+  ASSERT_OK(builder.Append(3));
+  ASSERT_OK_AND_ASSIGN(shared_ptr<arrow::Array> arr, builder.Finish())
```
lidavidm, line 31:
```suggestion
  ASSERT_OK_AND_ASSIGN(shared_ptr<arrow::Array> arr, builder.Finish());
```

cpp/code/common.h
```diff
@@ -0,0 +1,52 @@
...
+// under the License.
+
+#ifndef ARROW_COOKBOOK_COMMON_H
+#define ARROW_COOKBOOK_COMMON_H
+
+#include <sstream>
+#include <string>
+
+#include <arrow/result.h>
+#include <arrow/status.h>
```
lidavidm, line 25: FWIW, a Conda-installed copy of Arrow/C++ seems to contain arrow/testing/gtest_util.h so those could be used.

lidavidm (commented):
Just some thoughts on style/some minor typos I noticed.

cpp/CONTRIBUTING.md
westonpace: Yes, sorry, I didn't mean to imply it was part of the Arrow style guide.  I was trying to call it out as an exclusion.  The `std` namespace *should* be pretty well known and, if it is the only one we do a `using` for then it shouldn't be ambiguous (the examples don't live in a namespace).

But...it's also only 5 characters (although the default rendering only gives us 72) so it's not that big a deal.

cpp/code/CMakeLists.txt
westonpace: Yes, definitely.  Future PR though.  Should be easy enough to add later.  I'll create an issue.

cpp/CONTRIBUTING.md
lidavidm: Ah, ok. Yeah, I would feel it's best not to deviate from Arrow itself on this one. I feel like aliases for Arrow namespaces would probably save more space (ds = ::arrow::dataset and so on).

cpp/source/conf.py
westonpace: Added #43 to track this.

cpp/CONTRIBUTING.md
westonpace: Actually, this would ideally be `ARROW_RETURN_NOT_OK` to more accurately mirror real world usage.  I'll create an issue to tackle this.

westonpace: #44

cpp/code/common.h
westonpace: I think with #44 we may not need `ASSERT_*` variants at all.  However, we may want to create a recipe showing how to handle asserts in real world tests.  I'll add an issue.

build/index.html
westonpace: Done.

cpp/CONTRIBUTING.md
westonpace: Ok, I've removed `using std`.  At the moment things are ok.  I like the idea of arrow namespace aliases.  I think we can just play it by ear for now.

cpp/code/main.cc
westonpace: Nope.  Fixed.

westonpace:
@lidavidm I've addressed your suggestions.  Let me know if you have any further concerns.  Otherwise I'll merge this in tomorrow.

cpp/environment.yml
```diff
@@ -0,0 +1,115 @@
+name: cookbook-cpp
+channels:
+  - conda-forge
+dependencies:
+  - _libgcc_mutex=0.1=conda_forge
```
jorisvandenbossche, line 5: small nit, but might want to make this environment.yml file more generic? (only listing the actual packages needed (and not all their dependencies), and without the build hashes)

If you created this from a local conda environment you have, you can also use `conda env export --from-history` to get such a cleaner version.

cpp/CONTRIBUTING.md
```diff
@@ -0,0 +1,170 @@
...
+
+The above process requires conda to be installed and is primarily
+intended for build systems. See below for more information on setting
+up a development environment for developing recipes.
+
+# Developing C++ Recipes
+
+Every recipe is a combination of prose written in **reStructuredText**
+format using the `Sphinx <https://www.sphinx-doc.org/>`\_ documentation
+system and a snippet of a googletest test.
```
lidavidm, line 21: nits: if we spell out reStructuredText here, maybe we should do so above?

Should we link to reST and Googletest since we're also linking to Sphinx?

cpp/CONTRIBUTING.md
```diff
@@ -0,0 +1,170 @@
+# Bulding the C++ Cookbook
+
+The C++ cookbook combines output from a set of C++ test programs with
+an RST document tree rendered with sphinx.
```
lidavidm, line 4:
```suggestion
an RST document tree rendered with Sphinx.
```

cpp/CONTRIBUTING.md
```diff
@@ -0,0 +1,170 @@
...
+
+The entire document should serve as an example of how to use Arrow C++, not just the
+referenced snippets. This means that the below style rules and guidelines apply to
+source code that is not referenced by the cookbook itself.
+
+## Style
+
+This cookbook follows the same style rules as Arrow C++ which is the Google style
+guide with a few exceptions described
+[here](https://arrow.apache.org/docs/developers/cpp/development.html#code-style-linting-and-ci)
```
lidavidm, line 138: This should be changed to reST link syntax.

cpp/CONTRIBUTING.md
```diff
@@ -0,0 +1,170 @@
...
+in `main.cc` which runs after the tests run to dump these arrays to a .arrow file (i.e. the
+arrays will be serialized as a table using the Arrow IPC format).
+
+When the sphinx build runs the directive `recipe` (defined in `cpp/ext`) will be loaded.
+During this load the dataset of test outputs will be read. These test outputs will be used
+whenever a recipe is referenced.
+
+---
+
+All participation in the Apache Arrow project is governed by the Apache
```
lidavidm, line 168: Maybe we should call this out under its own heading?

lidavidm (commented):
Thanks, this looks great. I left some minor nits about wording/style but feel free to accept or ignore them.

cpp/environment.yml
westonpace: That was intentional, although maybe a little odd for the python world :).  It is often done in JS (e.g. npm/yarn) where they keep both the `package.xml` (in this case `requirements.txt`) and `package-lock.xml`.  The lock file is used most of the time but the package file is used when you want to upgrade to the latest version of all dependencies.  The purpose is two-fold:

1. Don't allow CI to fail because a downstream package pushed an update
2. Speed up CI because there is no need to run the solver.

The downside then is that you aren't always running against the latest and greatest and so you might miss an integration issue but, since these are only used for building the site (e.g. I'm not building a library), I don't think it's a big deal.

Things are a little trickier it seems with python because dependencies can have multiple distributions (JS dependencies that require native support build on install time) but as long as we're only using this in CI then I think it should be ok.

cpp/CONTRIBUTING.md
westonpace: This is a markdown file.

lidavidm: …doh, sorry.

lidavidm: Wait, then isn't the link above in reST syntax…?

westonpace: Good catch.  This doc was originally written in RST and then I converted it to markdown and missed a few things it seems.

PR #24, opened 2021-08-07
Disable jekyll

westonpace:
So it seems gh-pages was silently running jekyll on our gh-pages output which had no effect except that it ignores folders that start with `_` which is exactly what Sphinx likes to do when storing style pages.

Adding an empty `.nojekyll` at the root of the gh-pages is the way to fix this it seems.


thisisnic (approved):
LGTM!

Issue #23, opened 2021-08-07
Separate CI jobs for Python and R cookbook implementations


thisisnic:
Currently the cookbooks are built via `make all`; however, they have different dependencies and if no changes have been made to a particular implementation, this makes builds take longer.  The C++ build currently has its own CI job (see #22 ) - we could improve build/deploy time by doing the same for the Python/R cookbooks.

PR #26, opened 2021-08-10
Change gh-pages deploy to a single commit

westonpace:
Right now the deploy operation checks out the current gh-pages branch, does a `git add *` to update it, and then pushes it.  However, this has some downsides, most of which could be worked around with more complexity:

* If files are removed they won't disappear from gh-pages.
* If binary files (e.g. images, test data) change often they can clutter up the git repo (this is kind of an argument against any impl of gh-pages though)
* If no files changed it can fail the build because the commit / push will fail

An alternative method is to just force push a brand new branch with a single commit every single time which is what this PR does.

### DIFF
--- a/.github/workflows/deploy_cookbooks.yml
+++ b/.github/workflows/deploy_cookbooks.yml
@@ -45,6 +45,7 @@ jobs:
         run: |
           git config --global user.name 'GitHub Actions'
           git config --global user.email 'actions@github.com'
+          git checkout --orphan gh-pages-tmp
           git add *
           git commit -m 'publish built book'
-          git push origin gh-pages
+          git push -f origin gh-pages-tmp:gh-pages


jorisvandenbossche (approved):
+1, I don't think the history in the gh-pages branch is important?

thisisnic:
This looks good and solves the specific problem in a much cleaner way than before, but I think this might not be what we want to do in the context of other issues like #23 

In #23 , I'm looking at splitting out the CI into jobs that are dependent on which folder has been updated, so we're not attempting to rebuild all cookbooks when changes has been made to just one.  I imagine that as more implementations are added, this would increasingly slow down our CI.

I'm still working out how exactly to do this, but I think a solution involves having separate CI jobs for cookbook builds which are only triggered when that cookbook version has been updated, and then just those changes are copied to the relevant subdirectory of `build` and then pushed to the gh-pages branch.  I think we'd need to do something like `git add --all` to make sure we pick up any files which are removed, and `git commit --allow-empty` to allow empty commits (even though the jobs would be path-dependent, updates to tests could trigger a build without producing any output)

We'd still have to have a think about how to deal with binary files though - I did think maybe git lfs would be a possible solution but apparently it doesn't work with GH pages.

PR #28, opened 2021-08-10
Publish to gh-pages

thisisnic:
Currently gh-pages isn't working on this repo.  Taking a look at the Apache Airflow config file, we may have to add the publish field.

See: https://github.com/apache/airflow-site/blob/main/.asf.yaml

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -26,4 +26,5 @@ github:
   collaborators:
     - thisisnic
     - amol-
-
+  publish:
+    whoami:   gh-pages


westonpace:
I think that is how https://airflow.apache.org/ is update.  We need to be careful to not overwrite https://arrow.apache.org so I'm a little nervous with this PR.

westonpace:
I will note that there does not appear to be a https://apache.github.io/airflow

thisisnic:
Had another look at the docs and I think we wouldn't overwrite anything:

> By default, web sites will be published at $project.apache.org where $project is the sub-domain name of your project, as determined by the repository name.

Perhaps no projects use `.github.io` given there may not be a need/demand for this?  I'll close this ticket as I don't think we are wanting to publish to arrow-cookbook.apache.org

PR #30, opened 2021-08-10
Add Apache publishing / hosting

westonpace:
Very similar to #28 but adds the subdir.

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -31,3 +31,7 @@ notifications:
   commits:      commits@arrow.apache.org
   issues:       github@arrow.apache.org
   pullrequests: github@arrow.apache.org
+
+publish:
+  whoami: asf-site
+  subdir: cookbook
--- a/.github/workflows/deploy_cookbooks.yml
+++ b/.github/workflows/deploy_cookbooks.yml
@@ -32,7 +32,7 @@ jobs:
     runs-on: ubuntu-latest
     needs: make_cookbooks
     steps:
-      - name: Checkout gh-pages branch
+      - name: Checkout repo
         uses: actions/checkout@v2
         with:
           ref: gh-pages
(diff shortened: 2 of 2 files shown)


westonpace:
Leaving in draft to wait for feedback from ML

thisisnic:
Just wondering - do you think there anything else we would need to do to make this change publish to arrow.apache.org/cookbook instead of arrow-cookbook.apache.org/cookbook?

westonpace:
I don't think so but I agree the docs are vague.  I see your interpretation in #28 ...

> By default, web sites will be published at $project.apache.org where $project is the sub-domain name of your project, as determined by the repository name.

The "as determined by the repository name" and "$project is the sub-domain" certainly makes it seem like it would be arrow-cookbook.

However, `arrow-site` publishes to `arrow` so that is confusing.  I could believe that maybe the repo name is used to lookup the project name and the project name is used for publishing.

I could also believe that `arrow-site` has some legacy configuration settings from pre-`.asf.yaml` days and your interpretation is indeed correct.

So my hope is to start with this PR, see what happens, and if it is wrong, iterate from there.

westonpace:
Ah, when in doubt, refer to the source :)

https://github.com/apache/infrastructure-puppet/blob/e3dbb3c947dd1191b91ccd99e0f690802bff5b4c/modules/gitbox/files/asfgit/asfyaml.py#L591

The regex `re.match(r"(?:incubator-)?([^-.]+)", cfg.repo_name)` strips off any leading `incubator-` and then grabs everything up to the first `-`.  So `arrow-cookbook` will resolve to `arrow`.

thisisnic:
Wow, nice find! Never would have guessed that from the docs! 😬

westonpace:
No activity on the ML.  I'm going to try enabling this over the weekend so hopefully the impact is minimal if anything breaks.

westonpace:
Well, it doesn't seem that anything broke but the site also isn't hosted.  I'll follow up with Infra tomorrow.

PR #32, opened 2021-08-11
Move tests into implementation-specific folders and workflows

### LINKED ISSUE #23: Separate CI jobs for Python and R cookbook implementations

thisisnic:
Fixes #23 
This PR ensures that only implementation-specific tests are run on pull requests.  It also cancels any in-progress run of a workflow for the same PR.


westonpace (approved):
Looks good, should help cut down on check times.

PR #34, opened 2021-08-12
Enable the R cookbook to build against different Arrow versions and put in appropriate directory

thisisnic:


thisisnic:
This draft PR adds the functionality to specify Arrow versions in the Makefile and appears to work but there is still more work to be done around updating the CI jobs to supply the version number

PR #47, opened 2021-08-19
Writing Partitioned Datasets recipe for Python

### LINKED ISSUE #46: [Python] Add a recipe on how to save partitioned datasets to the Cookbook
We show how to read paritioned data, but we don't show how to write it in the Cookbook

amol-:
Fixes #46

### DIFF
--- a/python/source/io.rst
+++ b/python/source/io.rst
@@ -217,6 +217,63 @@ provided to :func:`pyarrow.csv.read_csv` to drive
     col1: int64
     ChunkedArray = 0 .. 99
 
+Writing Partitioned Datasets 
+============================
+
+When your dataset is big it usually makes sense to split it into
+multiple separate files. You can do this manually or use 
+:func:`pyarrow.dataset.write_dataset` to let Arrow do the effort
+of splitting the data in chunks for you.
+
+The ``partitioning`` argument allows to tell :func:`pyarrow.dataset.write_dataset`
+for which columns the data should be split. For example given a table
+with 100 numbers, we could add a ``chunk`` column that groups numbers
+in chunks of 10 numbers each.
+
+.. testcode::
+
+    data = pa.table({"numbers": range(100), 
+                     "chunk": [x // 10 for x in range(100)]})
+
+Then we could partition the data by the chunk column so that it
+gets saved in 10 different files:
+
+.. testcode::
+
+    import pyarrow as pa
+    import pyarrow.dataset as ds
+
+    ds.write_dataset(data, "./partitioned", format="parquet",
+                     partitioning=ds.partitioning(pa.schema([("chunk", pa.int8())])))
+
+Arrow will partition datasets in subdirectories by default, which will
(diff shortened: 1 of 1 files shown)


python/source/io.rst
```diff
@@ -217,6 +217,63 @@ provided to :func:`pyarrow.csv.read_csv` to drive
...
+
+Arrow will partition datasets in subdirectories by default, which will
+result in 10 different directories named with the value of the partitioning
+column and with file containing the data partition inside:
+
+.. testcode::
+
+    from pyarrow import fs
+
+    s3 = fs.LocalFileSystem()
```
jorisvandenbossche, line 257:
```suggestion
    localfs = fs.LocalFileSystem()
```

? (or don't import `fs`, and use `fs` as the name for the variable)

jorisvandenbossche (commented):
I understand that you are using the dummy "chunk" column to be able to have a partitioned dataset without having a natural column to partition on, but my feeling is that this is a bit of a workaround because we don't give a better alternative (yet). 

Personally, I would maybe use a more useful partitioning column as the main example, and then mention the "chunk" column as a way to do it if there is no column already present that can be used for partitioning (and then we can easily update this when a better API becomes available)

python/source/io.rst
amol-: ops, thanks for catching this one. I copy/pasted another block of code and ended up with the old variable name

amol-:
> I understand that you are using the dummy "chunk" column to be able to have a partitioned dataset without having a natural column to partition on, but my feeling is that this is a bit of a workaround because we don't give a better alternative (yet).
> 
> Personally, I would maybe use a more useful partitioning column as the main example, and then mention the "chunk" column as a way to do it if there is no column already present that can be used for partitioning (and then we can easily update this when a better API becomes available)

Switched to a real column (`year` of birthday) instead of using the chunk column.

amol-:
@jorisvandenbossche can you re-review? Should have addressed your concerns

python/source/io.rst
```diff
@@ -217,6 +217,65 @@ provided to :func:`pyarrow.csv.read_csv` to drive
...
+
+    import pyarrow as pa
+    import pyarrow.dataset as ds
+
+    ds.write_dataset(data, "./partitioned", format="parquet",
+                     partitioning=ds.partitioning(pa.schema([("year", pa.int16())])))
+
+Arrow will partition datasets in subdirectories by default, which will
+result in 10 different directories named with the value of the partitioning
+column and with file containing the data partition inside:
```
jorisvandenbossche, line 253: the "and with file containing the data partition inside" reads a bit strange. Maybe something like "each with a file containing the subset of the data for that partition"

jorisvandenbossche (commented):
Looks good! One small language remark

python/source/io.rst
amol-: :+1: reworded

jorisvandenbossche (approved):
Ready for merge?

PR #49, opened 2021-08-27
Recipe to read line delimited json as of ARROW-13708

amol-:
https://issues.apache.org/jira/browse/ARROW-13708

### DIFF
--- a/python/source/io.rst
+++ b/python/source/io.rst
@@ -497,3 +497,39 @@ the parquet file as :class:`ChunkedArray`
     pyarrow.Table
     col1: int64
     ChunkedArray = 0 .. 99
+
+Reading Line Delimited JSON
+===========================
+
+Arrow has builtin support for line-delimited JSON.
+Each line represents a row of data as a JSON object.
+
+Given some data in a file where each line is a JSON object
+containing a row of data:
+
+.. testcode::
+
+    import tempfile
+
+    with tempfile.NamedTemporaryFile(delete=False, mode="w+") as f:
+        f.write('{"a": 1, "b": 2.0, "c": 1}\n')
+        f.write('{"a": 3, "b": 3.0, "c": 2}\n')
+        f.write('{"a": 5, "b": 4.0, "c": 3}\n')
+        f.write('{"a": 7, "b": 5.0, "c": 4}\n')
+
+The content of the file can be read back to a :class:`pyarrow.Table` using
+:func:`pyarrow.json.read_json`
+
+.. testcode::
+
+    import pyarrow.json as pj
+
+    table = pj.read_json(f.name)
+
+.. testcode::
+
+    print(table.to_pydict())
+
+.. testoutput::
+
+    {'a': [1, 3, 5, 7], 'b': [2.0, 3.0, 4.0, 5.0], 'c': [1, 2, 3, 4]}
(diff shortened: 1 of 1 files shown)


python/source/io.rst
```diff
@@ -497,3 +497,39 @@ the parquet file as :class:`ChunkedArray`
...
     ChunkedArray = 0 .. 99
+
+Reading Line Delimited JSON
+===========================
+
+Arrow has builtin support for line-delimited JSON.
+Each line represents a row of data as a JSON object.
+
+Given some data in a file where each line is a JSON object
+containing a row of data:
```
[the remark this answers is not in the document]
jorisvandenbossche: If we need to make a choice, I am in favor of using a colon

python/source/io.rst
```diff
@@ -497,3 +497,39 @@ the parquet file as :class:`ChunkedArray`
...
+        f.write('{"a": 3, "b": 3.0, "c": 2}\n')
+        f.write('{"a": 5, "b": 4.0, "c": 3}\n')
+        f.write('{"a": 7, "b": 5.0, "c": 4}\n')
+
+The content of the file can be read back to a :class:`pyarrow.Table` using
+:func:`pyarrow.json.read_json`
+
+.. testcode::
+
+    import pyarrow.json as pj
```
jorisvandenbossche, line 525: Do we use the `pj` abbreviation anywhere else? If not, I would personally not introduce it, but either use `import pyarrow.json` (with `pa.json.read_json`), `from pyarrow import json` (with `json.read_json`) or `from pyarrow.json import read_json`. (I don't know if we are consistent with one of those patterns for other modules?)

(I know we have `pq` for parquet used widely, and that's probably too late to change, but I would personally not mimic this for all our modules)

python/source/io.rst
```diff
@@ -497,3 +497,39 @@ the parquet file as :class:`ChunkedArray`
...
+
+    table = pj.read_json(f.name)
+
+.. testcode::
+
+    print(table.to_pydict())
+
```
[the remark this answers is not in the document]
amol-: They were separated on purpose, mostly because the first code block is the real recipe, while the second is just to show the result.

python/source/io.rst
amol-: renamed `pj` to `pa.json`

python/source/io.rst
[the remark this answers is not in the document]
amol-: You are right, I didn't notice. I'm not sure many readers will notice but I added the colon to the second code block too.

amol-:
@westonpace @jorisvandenbossche I should have incorporated your feedback, I think this should be good to go

python/source/io.rst
```diff
@@ -497,3 +497,40 @@ the parquet file as :class:`ChunkedArray`
...
+.. testcode::
+
+    import tempfile
+
+    with tempfile.NamedTemporaryFile(delete=False, mode="w+") as f:
+        f.write('{"a": 1, "b": 2.0, "c": 1}\n')
+        f.write('{"a": 3, "b": 3.0, "c": 2}\n')
+        f.write('{"a": 5, "b": 4.0, "c": 3}\n')
+        f.write('{"a": 7, "b": 5.0, "c": 4}\n')
```
pitrou, line 518: Should the examples sometimes show null entries? I see none of them in the rest of the cookbook.

amol-: I think that when nulls change the behaviour of something and you need special code to deal with them we should have dedicated recipes. But in this case having nulls in the data shouldn't change how you deal with it, so I don't think there is much value in adding nulls

pitrou: Fair enough. I think it would be useful at some point to showcase null entries, just to show that they exist (silly, but some people may not be aware of that), but that's not necessarily in this section :-)

PR #51, opened 2021-08-27
Creating tables recipe

amol-:
https://issues.apache.org/jira/browse/ARROW-13715

### DIFF
--- a/python/source/create.rst
+++ b/python/source/create.rst
@@ -10,7 +10,7 @@ Tensors and all other Arrow entities.
 Creating Tables
 ===============
 
-Arrow supports tabular data in :class:`pyarrow.Table`, each column
+Arrow supports tabular data in :class:`pyarrow.Table`: each column
 is represented by a :class:`pyarrow.Array` and tables can be created
 by pairing multiple arrays with names for their columns
 
--- a/python/source/create.rst
+++ b/python/source/create.rst
@@ -19,9 +19,9 @@ by pairing multiple arrays with names for their columns
     import pyarrow as pa
 
     table = pa.table([
-      pa.array([1, 2, 3, 4, 5]),
-      pa.array(["a", "b", "c", "d", "e"]),
-      pa.array([1.0, 2.0, 3.0, 4.0, 5.0])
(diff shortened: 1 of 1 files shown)


amol-:
In the Python cookbook at the moment no subheadings were used, all recipes within a chapter were at the same level. We might consider adding further grouping but I think that should be done for all sections, not just this one, and I fear it might add confusion as we might end up with many subgroups of just two entries each.

python/source/create.rst
```diff
@@ -7,6 +7,32 @@ Tensors and all other Arrow entities.
 
 .. contents::
 
+Creating Tables
+===============
+
+Arrow supports tabular data in :class:`pyarrow.Table`, each column
```
jorisvandenbossche, line 13:
```suggestion
Arrow supports tabular data in :class:`pyarrow.Table`: each column
```

python/source/create.rst
```diff
@@ -7,6 +7,32 @@ Tensors and all other Arrow entities.
...
+is represented by a :class:`pyarrow.Array` and tables can be created
+by pairing multiple arrays with names for their columns
+
+.. testcode::
+
+    import pyarrow as pa
+
+    table = pa.table([
+      pa.array([1, 2, 3, 4, 5]),
```
jorisvandenbossche, line 22: Small nitpick, but use standard 4-space indentation here?

amol-: addressed

amol-:
I integrated suggestions, so this should now be good to go

python/source/create.rst
```diff
@@ -7,6 +7,32 @@ Tensors and all other Arrow entities.
 
 .. contents::
 
+Creating Tables
+===============
+
+Arrow supports tabular data in :class:`pyarrow.Table`: each column
+is represented by a :class:`pyarrow.Array` and tables can be created
```
pitrou, line 14: Internally, each column is a chunked array. But (non-chunked) arrays are also supported for constructing tables, for convenience.

amol-: :+1: replaced `Array` with `ChunkedArray`

Issue #44, opened 2021-08-19
Use `ARROW_RETURN_NOT_OK` (and friends) instead of `ASSERT_OK` (and friends) in recipes


westonpace:
Recipes should show how the code will be used in the wild.  `ASSERT_OK` is tailored for tests.  We could probably do this by putting recipes in their own function (possibly a part of the macro?) and then asserting the function returns ok.

PR #52, opened 2021-08-27
Use ARROW_RETURN_NOT_OK (and friends) instead of ASSERT_OK (and friends) in recipes

westonpace:
Addresses #44

### DIFF
--- a/cpp/CONTRIBUTING.md
+++ b/cpp/CONTRIBUTING.md
@@ -56,37 +56,17 @@ Each snippet source file contains a set of
 [googletest](https://github.com/google/googletest) tests. Feel free to
 use any googletest features needed to help setup and verify your test.
 To reference a snippet you need to surround it in `BeginRecipe` and
-`EndRecipe` calls. All snippets should be included in a call to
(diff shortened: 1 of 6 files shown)


pitrou:
Can you update cpp/CONTRIBUTING.md for the recipe API changes?

pitrou:
Ping @westonpace

westonpace:
Thanks for the ping.  I haven't had much time for cookbook work recently.  I'll take a look at this one on Friday and either fix it up or close it.

westonpace:
I updated the instructions and ended up changing things a bit further to the dataset new recipes that need to share state between recipes.

westonpace:
Ping @pitrou :)

pitrou (approved):

westonpace:
I came back to this and realized that @lidavidm 's flight test offers a much better solution.  We can simply move the test body into it's own method that returns a status.  This avoids the complexity of "recipe collections" and, since these tests should be quite small, there is little downside.  So I moved the tests to this approach instead.

lidavidm:
It was @pitrou's suggestion :sweat_smile:

PR #63, opened 2021-09-02
ARROW-13717: Creating arrays recipe

amol-:
https://issues.apache.org/jira/browse/ARROW-13717

### DIFF
--- a/python/source/create.rst
+++ b/python/source/create.rst
@@ -7,6 +7,68 @@ Tensors and all other Arrow entities.
 
 .. contents::
 
+Creating Arrays
+===============
+
+Arrow keeps data in continuous arrays optimised for memory footprint
+and SIMD analyses. In Python it's possible to build :class:`pyarrow.Array`
+starting from Python ``lists`` (or sequence types in general),
+``numpy`` arrays and ``pandas`` Series.
+
+.. testcode::
+
+    import pyarrow as pa
+
+    array = pa.array([1, 2, 3, 4, 5])
+
+.. testcode::
+
+    print(array)
+
+.. testoutput::
+
+    [
+      1,
+      2,
+      3,
+      4,
+      5
+    ]
+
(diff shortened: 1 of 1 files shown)


python/source/create.rst
```diff
@@ -7,6 +7,68 @@ Tensors and all other Arrow entities.
...
+
+    [
+      1,
+      2,
+      3,
+      4,
+      5
+    ]
+
+Arrays can also provide a ``mask`` to specify which values should
```
westonpace, line 38: I don't know if it's worth mentioning but the `mask` must be a numpy array (e.g. typical python list won't work)

westonpace (approved):
Some nits, but overall seems good to me.

python/source/create.rst
amol-: It's something I struggled a bit with, I wasn't sure if I should have explicitly mentioned that or not. Mostly for the reason that I would like to expand that to support ``pyarrow.Array`` too as masks ( https://issues.apache.org/jira/browse/ARROW-13883 ). So in the end I thought about not mentioning the constraint so that we don't leave behind a wrong information when the constraint is lightened

python/source/create.rst
```diff
@@ -7,6 +7,68 @@ Tensors and all other Arrow entities.
...
+    ]
+
+When building arrays from ``numpy`` or ``pandas``, Arrow will leverage
+optimized code paths that rely on the internal in-memory representation
+of the data by ``numpy`` and ``pandas``
+
+.. testcode::
+
+    import numpy as np
+    import pandas
```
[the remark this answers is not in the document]
amol-: :+1:

python/source/create.rst
westonpace: If you're planning on expanding support then I agree it should not be mentioned here.

PR #66, opened 2021-09-02
ARROW-13716: Add RecordBatch recipe

amol-:
https://issues.apache.org/jira/browse/ARROW-13716

### DIFF
--- a/python/source/create.rst
+++ b/python/source/create.rst
@@ -80,9 +80,9 @@ of a table.
     import pyarrow as pa
 
     batch = pa.RecordBatch.from_arrays([
-        pa.array([1, 2, 3, 4, 5]),
-        pa.array([10, 20, 30, 40, 50])
-    ], names=["first", "second"])
+        pa.array([1, 3, 5, 7, 9]),
+        pa.array([2, 4, 6, 8, 10])
+    ], names=["odd", "even"])
 
 Multiple batches can be combined into a table using 
 :meth:`pyarrow.Table.from_batches`
@@ -90,9 +90,9 @@ Multiple batches can be combined into a table using
 .. testcode::
 
     second_batch = pa.RecordBatch.from_arrays([
-        pa.array([6, 7, 8, 9, 10]),
-        pa.array([60, 70, 80, 90, 100])
-    ], names=["first", "second"])
+        pa.array([11, 13, 15, 17, 19]),
+        pa.array([12, 14, 16, 18, 20])
+    ], names=["odd", "even"])
 
     table = pa.Table.from_batches([batch, second_batch])
 
@@ -103,8 +103,8 @@ Multiple batches can be combined into a table using
 .. testoutput::
 
     pyarrow.Table
-    first: int64
-    second: int64
+    odd: int64
+    even: int64
 
 Equally, :class:`pyarrow.Table` can be converted to a list of 
 :class:`pyarrow.RecordBatch` using the :meth:`pyarrow.Table.to_batches`
--- a/python/source/create.rst
+++ b/python/source/create.rst
@@ -67,11 +67,11 @@ from a variety of inputs, including plain python objects
     :func:`pyarrow.array` for conversion to Arrow arrays,
     and will benefit from zero copy behaviour when possible.
 
-Creating RecordBatches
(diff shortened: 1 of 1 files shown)


python/source/create.rst
```diff
@@ -67,6 +67,58 @@ from a variety of inputs, including plain python objects
     :func:`pyarrow.array` for conversion to Arrow arrays,
     and will benefit from zero copy behaviour when possible.
 
+Creating RecordBatches
```
westonpace, line 70: Nit: Given the titles are prose I think it might be more correct to separate the words.
```suggestion
Creating Record Batches
```

python/source/create.rst
```diff
@@ -67,6 +67,58 @@ from a variety of inputs, including plain python objects
...
+======================
+
+Most IO operations in Arrow happen shipping batches of data
```
westonpace, line 73:
```suggestion
Most I/O operations in Arrow happen when shipping batches of data
```

python/source/create.rst
```diff
@@ -67,6 +67,58 @@ from a variety of inputs, including plain python objects
...
+to the destination, :class:`pyarrow.RecordBatch` are the way
```
westonpace, line 74:
```suggestion
to the destination.  :class:`pyarrow.RecordBatch` is the way
```

python/source/create.rst
```diff
@@ -67,6 +67,58 @@ from a variety of inputs, including plain python objects
...
+of a table.
+
+.. testcode::
+
+    import pyarrow as pa
+
+    batch = pa.RecordBatch.from_arrays([
+        pa.array([1, 2, 3, 4, 5]),
+        pa.array([10, 20, 30, 40, 50])
+    ], names=["first", "second"])
```
westonpace, line 85: Minor nit: Since we are introducing the concept of record batches here it may be clearer to the user if the column names and values are meaningful.  As it stands it might not be immediately obvious that `first` corresponds to `[1, 2, 3, 4, 5]` and `second` corresponds to `[10, 20, 30, 40, 50]`.  However, if the example were real world data it may be more clear.

python/source/create.rst
```diff
@@ -67,6 +67,58 @@ from a variety of inputs, including plain python objects
...
+
+multiple batches can be combined into a table using
```
westonpace, line 87:
```suggestion
Multiple batches can be combined into a table using 
```

python/source/create.rst
```diff
@@ -67,6 +67,58 @@ from a variety of inputs, including plain python objects
...
+
+    print(table)
+
+.. testoutput::
+
+    pyarrow.Table
+    first: int64
+    second: int64
+
+Equally, :class:`pyarrow.Table` can be converted to a set of
```
westonpace, line 109:
```suggestion
Equally, :class:`pyarrow.Table` can be converted to a list of 
```

westonpace (changes requested):
Looks pretty good.  Just a few grammatical/wording suggestions.

python/source/create.rst
amol-: Refactored the example to use odd and even numbers, should be immediately clear to the reader of things map

python/source/create.rst
```diff
@@ -67,6 +67,58 @@ from a variety of inputs, including plain python objects
     :func:`pyarrow.array` for conversion to Arrow arrays,
     and will benefit from zero copy behaviour when possible.
 
+Creating Record Batches
+======================
+
+Most I/O operations in Arrow happen when shipping batches of data
+to the destination.  :class:`pyarrow.RecordBatch` is the way
```
thisisnic, line 74:
```suggestion
to their destination.  :class:`pyarrow.RecordBatch` is the way
```

python/source/create.rst
```diff
@@ -67,6 +67,58 @@ from a variety of inputs, including plain python objects
...
+Arrow represents batches of data, they can be seen as a slice
```
thisisnic, line 75:
```suggestion
Arrow represents batches of data.  A RecordBatch can be seen as a slice
```

thisisnic (approved):
Tiny rephrasing suggestions, otherwise LGTM

PR #67, opened 2021-09-03
[R] - Schemas recipes

thisisnic:
ARROW-13722 and ARROW-13726 - Specifying schemas, and combining schemas


r/content/specify_data_types_and_schemas.Rmd
```diff
@@ -0,0 +1,205 @@
+# Defining Data Types
+
+As discussed in previous chapters, Arrow automatically handles the conversion of objects from native R data types to Arrow data types.  
+However, you might want to manually define data types, for example, to ensure interoperability with databases and data warehouse systems.
```
westonpace, line 4: Nit: This sentence doesn't quite work for me.  Even if you choose the data type Arrow will still need to convert from a native R type to an Arrow type.  Maybe something that uses the word "inference" or "picking the best data type".  For example:

```
Data in Arrow can be represented by a number of different data types.  When importing data from R the default behavior will pick the Arrow data type that is the safest match for the incoming R type.  However, you might want...
```

westonpace: Although even that example seems to fall short as description for the entire chapter because we should probably point out that this inference happens both when importing from R and when importing from files (or importing from any non-Arrow source really).

r/content/specify_data_types_and_schemas.Rmd
```diff
@@ -0,0 +1,205 @@
...
+
+## Specify data types when creating an Arrow table from an R object
+
+### Problem
```
westonpace, line 8: This `Problem`/`Solution` formatting is inconsistent with the other cookbook chapters.

r/content/specify_data_types_and_schemas.Rmd
```diff
@@ -0,0 +1,205 @@
...
+  company = c("AMZN", "GOOG", "BKNG", "TSLA"),
+  price = c(3463.12, 2884.38, 2300.46, 732.39),
+  date = rep(as.Date("2021-09-02"), 4)
+)
+
+# write dataset to disk
+write_dataset(share_data, path = "shares")
+
+# define field names and types
+share_schema <- schema(
```
westonpace, line 63: Nit: I know you described above but I have to wonder if someone looking at just this example might think that a schema is always required when reading or writing data.  Maybe update the below comment to something like...

```
# read the dataset in, using the schema instead of inferring the type
```

r/content/specify_data_types_and_schemas.Rmd
```diff
@@ -0,0 +1,205 @@
...
+| list^2^                  | list       |
+| bit64::integer64         | int64      |
+| difftime                 | time32     |
+| vctrs::vctrs_unspecified | null       |
+
+^1^: If the character vector exceeds 2GB of strings, it will be converted to a 
+`large_utf8` Arrow type
+
+^2^: Only lists where all elements are the same type are able to be translated 
+to Arrow list type (which is a "list of" some type).
```
westonpace, line 114: It's a little odd that you call out `^1^` and `^2^` here but the above table also has `^7^` which you don't describe until further down.  Maybe change the `7` to `3` and describe it here or move all the footnotes to the lower section.

r/content/specify_data_types_and_schemas.Rmd
```diff
@@ -0,0 +1,205 @@
...
+
+The data types created via default mapping from R to Arrow are not the only ones
+which exist, and alternative Arrow data types may compatible with each R data 
+type.  The compatible data types are shown in the table below.
```
westonpace, line 118: This sentence doesn't read quite right.  "and alternative Arrow data types may *be* compatible..." perhaps?

r/content/specify_data_types_and_schemas.Rmd
```diff
@@ -0,0 +1,205 @@
...
+| boolean           | logical                      |
+| int8              | integer                      |
+| int16             | integer                      |
+| int32             | integer                      |
+| int64             | integer^3^                   |
+| uint8             | integer                      |
+| uint16            | integer                      |
+| uint32            | integer^3^                   |
+| uint64            | integer^3^                   |
+| float16           | -                            |
```
westonpace, line 133: What does `-` mean?  Does this mean the Arrow data type has no corresponding R type?  What happens in that case?  Is it a runtime error?  Can you expand on this below?

r/content/specify_data_types_and_schemas.Rmd
```diff
@@ -0,0 +1,205 @@
...
+| float32           | double                       |
+| float64 ^7^       | double                       |
+| utf8              | character                    |
```
westonpace, line 136: Nit: Would it perhaps be better to group `utf8` and `large_utf8` together in this table?

r/content/specify_data_types_and_schemas.Rmd
```diff
@@ -0,0 +1,205 @@
...
+| binary            | arrow_binary ^5^             |
+| fixed_size_binary | arrow_fixed_size_binary ^5^  |
+| date32            | Date                         |
+| date64            | POSIXct                      |
+| time32            | hms::difftime                |
+| time64            | hms::difftime                |
+| timestamp         | POSIXct                      |
+| duration          | -                            |
+| decimal           | double                       |
```
westonpace, line 145: If I were a naive user looking at this table then I'd probably be wondering "What is decimal and why does it also correspond to double?"  I'm not sure if that has to be answered here though.

r/content/specify_data_types_and_schemas.Rmd
```diff
@@ -0,0 +1,205 @@
...
+^5^: `arrow*_binary` classes are implemented as lists of raw vectors. 
+
+^6^: `arrow*_list` classes are implemented as subclasses of `vctrs_list_of` 
+with a `ptype` attribute set to what an empty Array of the value type converts to. 
+
+^7^: `float64` and `double` are the same concept and data type in Arrow C++; 
+however, only `float64()` is used in arrow as the function `double()` already exists in base R
+
+
+## Combine and harmonize schemas
```
westonpace, line 175:
```suggestion
## Combine and unify schemas
```
This line is prose so harmonize as an alias for unify could work but I fear a more technical minded reader might try and read into it more than they should and expect some formal definition of harmonizing schemas.

r/content/specify_data_types_and_schemas.Rmd
```diff
@@ -0,0 +1,205 @@
...
+
+```{r, combine_schemas}
+# create first schema to combine
+country_code_schema <- schema(country = utf8(), code = utf8())
+
+# create second schema to combine
+country_phone_schema <- schema(country = utf8(), phone_prefix = int8())
+
+# combine schemas
+combined_schemas <- unify_schemas(country_code_schema, country_phone_schema)
```
westonpace, line 194: This begs the question, "What happens if the types are different but the names are the same?"

r/content/specify_data_types_and_schemas.Rmd
thisisnic: Fair points, rephrased now.

r/content/specify_data_types_and_schemas.Rmd
thisisnic: I was planning on rewriting all of the other recipes with this format shortly in a follow-up PR.  My thinking here is that I'd rather get this merged first as I have multiple PRs for multiple chapters open and don't want to end up with too many conflicts.  Of course, that doesn't stop me adding this formatting to those PRs if they haven't been merged yet, which I will take a look at doing.

r/content/specify_data_types_and_schemas.Rmd
thisisnic: Can't object to suggestions that improve clarity; cheers, updated!

r/content/specify_data_types_and_schemas.Rmd
thisisnic: Yup, updated

r/content/specify_data_types_and_schemas.Rmd
thisisnic: That's a good question; I think it belongs in the function docs rather than the cookbook, but is definitely worth answering so I've expanded this ticket to also cover it: https://issues.apache.org/jira/browse/ARROW-13886

r/content/specify_data_types_and_schemas.Rmd
westonpace: As long as there is an issue/JIRA for the format update (so we don't lose track of it) I think that's fine.

thisisnic: I've opened #68 for this now.

r/content/specify_data_types_and_schemas.Rmd
thisisnic: Updated now to be separate tables with separate numbering systems as it's much cleaner

r/content/specify_data_types_and_schemas.Rmd
thisisnic: Have added an example and link to it from appendix. Still needs to work out exactly which section it should go in though.

r/content/specify_data_types_and_schemas.Rmd
thisisnic: Good thinking

r/content/specify_data_types_and_schemas.Rmd
thisisnic: Yep, moved

r/content/specify_data_types_and_schemas.Rmd
thisisnic: Good question - have added an example on how this causes an error!

r/content/specify_data_types_and_schemas.Rmd
```diff
@@ -0,0 +1,279 @@
...
+})
+```
+```{r, include=FALSE}
+unlink("oscars_data", recursive = TRUE)
+```
+
+## Combine and unify schemas
+
+You have a dataset split across multiple sources for which you have separate 
+schemas that you want to combine.
```
westonpace, line 201: I'm a bit hesitant on the word "dataset" here.  A dataset (as in datasets API) will already unify schemas for you (well, it should, it doesn't always work as well as it could).

I think I'm just hesitant on the recipe in general.  I can't come up with a good use case where you'd want to combine completely unrelated schemas.  I think this recipe is superfluous given the next recipe.

r/content/specify_data_types_and_schemas.Rmd
```diff
@@ -0,0 +1,279 @@
...
+    schema(country = utf8(), code = utf8(), longitude = float32(), latitude = float32())
+  )
+})
+```
+
+
+### Discussion
+
+If you have two schemas containing fields with the same names and types, you can
+ combine them into a single field.
```
westonpace, line 232:
```suggestion
If you have two different schemas with some overlapping fields which have the same names and types, you can
 combine them into a single schema.
```

Nit: I think I get what you are after but off the cuff this wasn't clear to me.  I don't feel too strongly about it though.

westonpace (approved):
Great improvement, I still have a few nits, but I could be convinced it is fine how it is too.

r/content/specify_data_types_and_schemas.Rmd
jorisvandenbossche: In addition, this unifying will typically be done by `open_dataset(..)` (or can be done, by using the correct keyword argument). It might be more useful to show that?

r/content/specify_data_types_and_schemas.Rmd
```diff
@@ -0,0 +1,250 @@
...
+```{r, float_16_conversion, error=TRUE, eval=FALSE}
+oscars <- tibble::tibble(
+  actor = c("Katharine Hepburn", "Meryl Streep", "Jack Nicholson"),
+  num_awards = c(4, 3, 3)
+)
+
+# Set up schema with "num_awards" as float16 which doesn't have an R equivalent
+oscars_schema <- schema(actor = string(), num_awards = float16())
+
+Table$create(oscars, schema = oscars_schema)
```
jorisvandenbossche, line 117: You are passing a schema to the create method (so used in the R->arrow conversion, not when casting an Arrow table), which is only shown below. So maybe move this discussion to the topic below?

r/content/specify_data_types_and_schemas.Rmd
```diff
@@ -0,0 +1,250 @@
...
+```{r, include=FALSE}
+unlink("oscars_data", recursive = TRUE)
+```
+
+### Discussion
+
+If your dataset contains files with different schemas with some overlapping 
+fields which have the same names and types, they will be combined by
+`open_dataset()` automatically.  The code below demonstrates how this works 
+via the `unify_schemas()` function.
```
jorisvandenbossche, line 203: Unfortunately it's actually not fully equivalent. `unify_schemas` will error if types are not the same for fields with the same name, while open_dataset can do some schema evolution (promoting some types). This might be too much a detail to go into here, though (and it's maybe also rather a missing feature in `unify_schemas`)

jorisvandenbossche (commented):
Looks good! Added two more comments

r/content/specify_data_types_and_schemas.Rmd
thisisnic: Sorry, please can you rephrase this? I don't know what you mean

thisisnic: Actually, never mind, I see now!

r/content/specify_data_types_and_schemas.Rmd
thisisnic: Thanks for pointing this out.  I have opened [ARROW-14190](https://issues.apache.org/jira/browse/ARROW-14190) to discuss this, and removed this section to the cookbook as I agree it's too much detail here.

r/content/specify_data_types_and_schemas.Rmd
thisisnic: I have now refactored this example to use casting instead of schema specification.

jorisvandenbossche (approved):

Issue #56, opened 2021-08-31
[Python] Make pytest fails


Nlte:
I'm getting 1 fail test when running the `make pytest` target.

```
Document: io
------------
**********************************************************************
File "io.rst", line 799, in default
Failed example:
    dataset = ds.dataset("s3://ursa-labs-taxi-data/2011",
                         partitioning=["month"])
    for f in dataset.files[:10]:
        print(f)
    print("...")
Exception raised:
    Traceback (most recent call last):
      File "/usr/local/Cellar/python@3.9/3.9.6/Frameworks/Python.framework/Versions/3.9/lib/python3.9/doctest.py", line 1336, in __run
        exec(compile(example.source, filename, "single",
      File "<doctest default[0]>", line 1, in <module>
        dataset = ds.dataset("s3://ursa-labs-taxi-data/2011",
      File "/Users/nathanael.leaute/Documents/github/arrow-cookbook/venv/lib/python3.9/site-packages/pyarrow/dataset.py", line 655, in dataset
        return _filesystem_dataset(source, **kwargs)
      File "/Users/nathanael.leaute/Documents/github/arrow-cookbook/venv/lib/python3.9/site-packages/pyarrow/dataset.py", line 410, in _filesystem_dataset
        return factory.finish(schema)
      File "pyarrow/_dataset.pyx", line 2402, in pyarrow._dataset.DatasetFactory.finish
      File "pyarrow/error.pxi", line 143, in pyarrow.lib.pyarrow_internal_check_status
      File "pyarrow/error.pxi", line 114, in pyarrow.lib.check_status
    OSError: Error creating dataset. Could not read schema from 'ursa-labs-taxi-data/2011/01/data.parquet': Could not open Parquet input source 'ursa-labs-taxi-data/2011/01/data.parquet': AWS Error [code 15]: Access Denied. Is this a 'parquet' file?
**********************************************************************
1 items had failures:
   1 of  27 in default
27 tests in 1 items.
26 passed and 1 failed.
***Test Failed*** 1 failures.

```

It seems like the ACL on the ursa-labs-taxi-data bucket doesn't allow public access. I don't know if you want to open up the bucket / prefix to the public and incur that aws bandwidth costs though. Those are definitely a thing.

westonpace:
[...]

PR #70, opened 2021-09-09
Adding anonymous flag to s3

drabastomek:
Without the `anonymous=True` flag I was not able to read data from the `ursa-labs-taxi-data` S3 bucket. I updated the S3 section so this should work every time.

### DIFF
--- a/python/source/io.rst
+++ b/python/source/io.rst
@@ -422,8 +422,7 @@ by ``month`` using
 
 .. testcode::
 
-    dataset = ds.dataset("2011", 
-                         filesystem=s3,
+    dataset = ds.dataset("s3://ursa-labs-taxi-data/2011",
                          partitioning=["month"])
     for f in dataset.files[:10]:
         print(f)
@@ -450,6 +449,27 @@ or :meth:`pyarrow.dataset.Dataset.to_batches` like you would for a local one.
 
     It is possible to load partitioned data also in the ipc arrow
     format or in feather format.
+
+.. warning::
+
+    If the above code throws an error most likely the reason is your
+    AWS credentials are not set. Follow these instructions to get
+    ``AWS Access Key Id`` and ``AWS Secret Access Key``: 
+    `AWS Credentials <https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_access-keys.html>`_.
+
+    The credentials are normally stored in ``~/.aws/credentials`` (on Mac or Linux)
+    or in ``C:\Users\<USERNAME>\.aws\credentials`` (on Windows) file. 
+    You will need to either create or update this file in the appropriate location.
+
+    The contents of the file should look like this:
+
+    .. code-block:: bash 
+
+        [default]
+        aws_access_key_id=<YOUR_AWS_ACCESS_KEY_ID>
+        aws_secret_access_key=<YOUR_AWS_SECRET_ACCESS_KEY>
+
+
 
 Write a Feather file
 ====================
--- a/python/source/io.rst
+++ b/python/source/io.rst
@@ -394,7 +394,10 @@ partitioned data coming from remote sources like S3 or HDFS.
     from pyarrow import fs
 
     # List content of s3://ursa-labs-taxi-data/2011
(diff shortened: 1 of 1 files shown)


python/source/io.rst
```diff
@@ -419,7 +422,8 @@ by ``month`` using
 
 .. testcode::
 
-    dataset = ds.dataset("s3://ursa-labs-taxi-data/2011", 
+    dataset = ds.dataset("2011", 
+                         filesystem=s3
```
westonpace, line 426: Hmm, I suppose I feel this changes the intent a little if the goal was to demonstrate the ability to create a dataset from a URI.

That being said, I don't know how to make this test work effectively with a URI.  If `anonymous` is not provided then credentials will be fetched from by the AWS SDK (e.g. from ~/.aws/credentials).  If no credentials at all are found then the request will fail.  If you configure any kind of credentials then the request will pass.  Perhaps the C++ library could catch the "no credentials found" error and try again with anonymous access.

westonpace (commented):
I like the idea as it has already come up once before.  It's a clear win for the case where we were explicitly using a filesystem already.  However, for the other recipe, I think we were intentionally demonstrating how to use a URI (instead of creating a filesystem).  If that is the case I'd leave the temperamental approach for now.  If that wasn't the intent of the recipe then I'm fine with both changes.  I don't feel too strongly either way.

@amol- @jorisvandenbossche

python/source/io.rst
drabastomek: @westonpace I understand the reasoning about the URI. Perhaps, instead of changing this to `filesystem` we do, as you suggest, add `~/.aws/credentials` as an explicit step before this code? I think this would make sure this code works even if credentials are not present.

drabastomek: @westonpace I changed the code back to using URI and added a warning that if this code fails most likely it's because of the AWS credentials not being set. Added instructions how to set these properly.

westonpace: Thanks, I'm good with this approach.  If amol- or jorisvandenbossche have better ideas when they get back we can fix it then.

python/source/io.rst
```diff
@@ -447,6 +450,27 @@ or :meth:`pyarrow.dataset.Dataset.to_batches` like you would for a local one.
     It is possible to load partitioned data also in the ipc arrow
     format or in feather format.
 
+.. warning::
+
+    If the above code throws an error most likely the reason is your
+    AWS credentials are not set. Follow these instructions to get
```
jorisvandenbossche, line 456: Based on https://github.com/apache/arrow-cookbook/issues/56, it seems you can also run into errors if you *have* set AWS credentials? (in which case adding anonymous=True can help, so it might still be worth mentioning it here in the warning?)

drabastomek: This would need to revert to adding `filesystem` parameter rather than a `anonymous` flag directly in the `dataset` object. Unless I'm doing something wrong.

jorisvandenbossche: Yeah, I wouldn't suggest changing the actual example again to use `filesystem` instead of a URI. But in the warning box it might still be worth mentioning the `anonymous` flag (and that you can then pass it with `filesystem=`), as based on my reading of #56, it might solve similar errors. 

Now, it's not really clear to me when what is a exactly giving problems. In #56 it's having credentials set that makes reading the public bucket not work, except if you use anonymous flag, while the explanation added in the PR above seems to indicate that _setting_ credentials can help.

drabastomek: Makes sense. I'll update the warning. 

However, looks like #56 was having the reverse of my problems: having the `credentials` file but still unable to connect to a public bucket. I wonder if the auth code on our end is buggy or the AWS throws...

westonpace: I experimented and it seems, for me at least, the S3 URL works if I have no credentials file (presumably it is anonymous in that case) and it works if I have credentials belonging to the Ursa organization in S3.  It fails if I have invalid credentials (not an interesting use case) and if I have valid credentials that aren't part of the Ursa organization (this is the case I think that people keep running into unexpectedly, which is aided by the anonymous=True flag)

PR #75, opened 2021-09-15
ARROW-13725: Unify schemas recipe

amol-:
https://issues.apache.org/jira/browse/ARROW-13725

### DIFF
--- a/python/source/schema.rst
+++ b/python/source/schema.rst
@@ -109,3 +109,86 @@ as far as they are compatible
     col1: int32
     col2: string
     col3: double
+
+Merging multiple schemas
+========================
+
+When you have multiple separate groups of data that you want to combine
+it might be necessary to unify their schemas to create a superset of them
+that applies to all data sources.
+
+.. testcode::
+
+    import pyarrow as pa
+
+    first_schema = pa.schema([
+        ("country", pa.string()),
+        ("population", pa.int32())
+    ])
+
+    second_schema = pa.schema([
+        ("country_code", pa.string()),
+        ("language", pa.string())
+    ])
+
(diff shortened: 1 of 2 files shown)


python/source/schema.rst
```diff
@@ -108,4 +108,87 @@ as far as they are compatible
...
+
+.. testoutput::
+
+    country: string
+    population: int32
+    country_code: string
+    language: string
+
+If the combined schemas have overlapping columns, they can still be combined
+as far as the colliding columns retain the same type (``country_code``):
```
jorisvandenbossche, line 151: Small suggestion (but take it or leave it :)): I would personally directly show an example where there is a column in common, instead of splitting it in two examples first without common column and then with. 
Having no columns in common at all doesn't seem like the typical use case for this, so directly showing the second example can simplify the cookbook entry a bit (one example less to show), while you can still explicitly mention how the common column is handled.

amol-: I actually went in the direction of showing the case where there are no overlapping columns first because I thought that when talking about "merging" two things it's easier to get what's going on if you see a perfect union of the two original collections.

jorisvandenbossche: Yes, but as far as I understood the unify function, its primary intended use case is to unify slightly different schemas (eg different ordering, missing column, ..), and not to merge schemas in a typical "join" context where you join two different tables (although this is of course also possible). 
So while the second use might be easier to explain in itself, IMO it doesn't really help overall to explain the typical use of `unify_schemas`. 
(but as I said, take it or leave it :), those things will always be inherently subjective (so I should maybe not have reacted again :))

PR #84, opened 2021-10-08
ARROW-13710: Arrow Flight recipe

amol-:
https://issues.apache.org/jira/browse/ARROW-13710

### DIFF
--- a/python/source/flight.rst
+++ b/python/source/flight.rst
@@ -6,8 +6,8 @@ Recipes related to leveraging Arrow Flight protocol
 
 .. contents::
 
-Using an Arrow Flight RPC server
-================================
+Simple Parquet storage service with Arrow Flight
+================================================
 
 Suppose you want to implement a service that can store, send and receive
 Parquet files using the Arrow Flight protocol, 
@@ -17,7 +17,6 @@ and particularly through the :class:`pyarrow.flight.FlightServerBase` class.
 .. testcode::
 
     import pathlib
-    import threading
 
     import pyarrow as pa
     import pyarrow.flight
@@ -120,6 +119,7 @@ Our server can then be started with
     # Code block to start for real a server in background
     # and wait for it to be available.
     # Previous code block is just to show to user how to start it.
+    import threading
     server = FlightServer()
     server._repo.mkdir(exist_ok=True)
     t = threading.Thread(target=server.serve)


python/source/flight.rst
```diff
@@ -0,0 +1,228 @@
...
+
+Recipes related to leveraging Arrow Flight protocol
+
+.. contents::
+
+Using an Arrow Flight RPC server
+================================
+
+Suppose you want to implement a service that can store, send and receive
+parquet files using the Arrow Flight protocol,
```
lidavidm, line 13:
```suggestion
Parquet files using the Arrow Flight protocol, 
```

python/source/flight.rst
```diff
@@ -0,0 +1,228 @@
...
+of actually fetching the exposed data streams and sending them to the client.
+
+Allowing to list and dowload data streams would be pretty useless if we didn't
+expose a way to create them, this is the responsability of
+:meth:`pyarrow.flight.FlightServerBase.do_put` which is in charge of receiving
+new data from the client and dealing with it (in this case saving it
+into a parquet file)
+
+This are the most common Arrow Flight requests, if we need to add more
+functionalities, we can do so using custom actions.
```
lidavidm, line 99: Is it worth noting somewhere that if not implemented, all Flight methods have a default implementation? 
Similarly, it may be worth linking to the list of all methods. (Oh, though the docs don't really have a good listing. That should be fixed.)

python/source/flight.rst
```diff
@@ -0,0 +1,228 @@
...
+        server = FlightServer()
+        server._repo.mkdir(exist_ok=True)
+        server.serve()
+
+.. testcode::
+    :hide:
+
+    # Code block to start for real a server in background
+    # and wait for it to be available.
+    # Previous code block is just to show to user how to start it.
```
lidavidm, line 122: In that case, should we move the `import threading` from the first code block to here?

python/source/flight.rst
```diff
@@ -0,0 +1,228 @@
...
+                yield self._make_flight_info(dataset.name)
+
+        def get_flight_info(self, context, descriptor):
+            return self._make_flight_info(descriptor.path[0].decode('utf-8'))
+
+        def do_put(self, context, descriptor, reader, writer):
+            dataset = descriptor.path[0].decode('utf-8')
+            dataset_path = self._repo / dataset
+            data_table = reader.read_all()
+            pa.parquet.write_table(data_table, dataset_path)
```
lidavidm, line 60: A follow-up example might be to do this in a streaming way to reduce memory pressure (ditto for do_get).

python/source/flight.rst
```diff
@@ -0,0 +1,228 @@
...
+Allowing to list and dowload data streams would be pretty useless if we didn't
+expose a way to create them, this is the responsability of
+:meth:`pyarrow.flight.FlightServerBase.do_put` which is in charge of receiving
+new data from the client and dealing with it (in this case saving it
+into a parquet file)
+
+This are the most common Arrow Flight requests, if we need to add more
+functionalities, we can do so using custom actions.
+
+In the previous example a `drop_dataset` custom action is added.
```
lidavidm, line 101:
```suggestion
In the previous example a ``drop_dataset`` custom action is added. 
```

python/source/flight.rst
```diff
@@ -0,0 +1,228 @@
...
+    import threading
+
+    import pyarrow as pa
+    import pyarrow.flight
+    import pyarrow.parquet
+
+
+    class FlightServer(pa.flight.FlightServerBase):
+
+        def __init__(self, location="grpc://0.0.0.0:8815",
```
lidavidm, line 29: As an alternative to hardcoding a port, you can bind to port 0, then print out `server.port` after creating the server instance (before calling serve() - confusingly, the server is started as soon as you instantiate the class)

python/source/flight.rst
```diff
@@ -0,0 +1,228 @@
+============
+Arrow Flight
+============
+
+Recipes related to leveraging Arrow Flight protocol
+
+.. contents::
+
+Using an Arrow Flight RPC server
```
lidavidm, line 9: This title is a little generic for what it shows (I would have expected much less content/a much more minimal example), perhaps it could be called something like "A simple Parquet storage service with Arrow Flight"?

amol-: Good suggestion, I struggled a bit with naming this recipe as it covers so many things as you pointed out.

python/source/flight.rst
amol-: Good suggestion, created https://github.com/apache/arrow-cookbook/issues/86

PR #87, opened 2021-10-13
ARROW-13712: Reading and Writing Compressed Data

amol-:
https://issues.apache.org/jira/browse/ARROW-13712

### DIFF
--- a/python/source/io.rst
+++ b/python/source/io.rst
@@ -638,7 +638,7 @@ like CSV, but have been compressed by an application.
 Reading compressed formats that have native support for compression
 doesn't require any special handling. We can for example read back
 the Parquet and Feather files we wrote in the previous recipe
-simply invoking :meth:`pyarrow.feather.read_table` and
+by simply invoking :meth:`pyarrow.feather.read_table` and
 :meth:`pyarrow.parquet.read_table`:
 
 .. testcode::


python/source/io.rst
```diff
@@ -577,4 +577,121 @@ The content of the file can be read back to a :class:`pyarrow.Table` using
...
 .. testoutput::
 
-    {'a': [1, 3, 5, 7], 'b': [2.0, 3.0, 4.0, 5.0], 'c': [1, 2, 3, 4]}
\ No newline at end of file
+    {'a': [1, 3, 5, 7], 'b': [2.0, 3.0, 4.0, 5.0], 'c': [1, 2, 3, 4]}
+
+Writing Compressed Data
+=======================
+
+Arrow provides support for writing files in compressed format,
```
thisisnic, line 585:
```suggestion
Arrow provides support for writing files in compressed formats,
```

python/source/io.rst
```diff
@@ -577,4 +577,121 @@ The content of the file can be read back to a :class:`pyarrow.Table` using
...
+both for formats that provide it natively like Parquet or Feather,
```
thisisnic, line 586:
```suggestion
both for formats that provide compression natively like Parquet or Feather,
```

python/source/io.rst
```diff
@@ -577,4 +577,121 @@ The content of the file can be read back to a :class:`pyarrow.Table` using
...
+and for formats that don't support it out of the box like CSV.
```
thisisnic, line 587:
```suggestion
and for formats that don't support compression out of the box like CSV.
```

python/source/io.rst
```diff
@@ -577,4 +577,121 @@ The content of the file can be read back to a :class:`pyarrow.Table` using
...
+
+Given a table:
+
+.. testcode::
+
+    table = pa.table([
+        pa.array([1, 2, 3, 4, 5])
+    ], names=["numbers"])
+
+Writing it compressed to parquet or feather requires passing the
```
thisisnic, line 597:
```suggestion
Writing it compressed to Parquet or Feather requires passing the
```

python/source/io.rst
```diff
@@ -577,4 +577,121 @@ The content of the file can be read back to a :class:`pyarrow.Table` using
...
+:func:`pyarrow.parquet.write_table` functions:
+
+.. testcode::
+
+    pa.feather.write_feather(table, "compressed.feather",
+                             compression="lz4")
+    pa.parquet.write_table(table, "compressed.parquet",
+                           compression="lz4")
+
+You can refer to the two functions documentation for a complete
```
thisisnic, line 608:
```suggestion
You can refer to each of those functions' documentation for a complete
```

python/source/io.rst
```diff
@@ -577,4 +577,121 @@ The content of the file can be read back to a :class:`pyarrow.Table` using
...
+list of supported compression formats.
+
+.. note::
+
+    Arrow actually uses compression by default when writing
+    parquet or feather files. Feather is compressed using ``lz4``
```
thisisnic, line 614:
```suggestion
    Parquet or Feather files. Feather is compressed using ``lz4``
```

python/source/io.rst
```diff
@@ -577,4 +577,121 @@ The content of the file can be read back to a :class:`pyarrow.Table` using
...
+This requires decompressing the file when reading it back,
+which can be done using :class:`pyarrow.CompressedInputStream`
+as explained in the next recipe.
+
+Reading Compressed Data
+=======================
+
+Arrow provides support for reading compressed files,
+both for formats that provide it natively like Parquet or Feather,
+and for formats that don't support it out of the box like CSV.
```
thisisnic, line 635: I was thinking this could be a tiny bit clearer, though I think the "external application" phrasing needs tweaking a bit.
```
Arrow provides support for reading compressed files,
both for formats that provide it natively like Parquet or Feather,
and for files in formats that don't support compression natively, 
like CSV, but have been compressed by an external application.
```

python/source/io.rst
jorisvandenbossche, line 597: For Parquet passing the argument is not actually "required", since we use compression by default (using snappy). You can still use the keyword to change the default compression strategy, though.

python/source/io.rst
```diff
@@ -577,4 +577,121 @@ The content of the file can be read back to a :class:`pyarrow.Table` using
...
+    print(table_parquet)
+
+.. testoutput::
+
+    pyarrow.Table
+    numbers: int64
+
+Reading data from formats that don't have native support for
+compression instead involves decompressing them before decoding them.
+This can be done using the :class:`pyarrow.CompressedInputStream` class
```
jorisvandenbossche, line 665: The `read_csv` docstring seems to indicate it will actually decompress by default if it recognizes the compressed file extension, so maybe we can show both ways?

python/source/io.rst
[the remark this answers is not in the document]
amol-: They are linked right before the code snippet, that's why I avoided relinking them again here

python/source/io.rst
amol-: :+1:

python/source/io.rst
amol-: I wanted to provide a general example on how to read compressed data in generic format, `CSV` is just an excuse that by chance provides the compression out of the box, but the same could apply for example to `JSON Line-Delimited` in which case there is builtin support for compression.

That's why I went for explicit usage of `CompressedInputStream` (which will work with whatever format) even thought the example writes a CSV file.

PR #109, opened 2021-12-01
[Python] Add Flight streaming example

### LINKED ISSUE #86: [Python][Flight] Recipe to show how to stream data in `do_get`
The current flight recipe loads in memory the whole content of the datasets before writing them to the response, it would be good to have a recipe that also shows how to stream data for the case of big datasets that might not fit in memory.

lidavidm:
Fixes #86.

### DIFF
--- a/python/source/flight.rst
+++ b/python/source/flight.rst
@@ -238,7 +238,6 @@ advantage of this:
 
    import pathlib
 
-   import numpy as np
    import pyarrow as pa
    import pyarrow.flight
    import pyarrow.parquet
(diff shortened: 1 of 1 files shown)


lidavidm:
<details>
<summary>Rendered</summary>

image

</details>

lidavidm:
CC @amol- if you have any comments here.

python/source/flight.rst
```diff
@@ -226,3 +226,198 @@ we might list all parquet files that are currently stored by the server:
...
+           with dataset_path.open("wb") as sink:
+               with pa.parquet.ParquetWriter(sink, reader.schema) as writer:
+                   for chunk in reader:
+                       writer.write_table(pa.Table.from_batches([chunk.data]))
+
+       def do_get(self, context, ticket):
+           dataset = ticket.ticket.decode('utf-8')
+
+           # Stream data from a generator we implement
+           if dataset == ":random:":
```
amol-, line 289: iI think that the `:random:` special dataset might be confusing for a reader in the context of this recipe. It's great for playing around, but the recipe goal was to read back datasets you stored, So a reader might wonder what's the purpose of that code block and get confused by it.

lidavidm: Makes sense. I wanted to demonstrate GeneratorStream, so instead I'll use it to replace RecordBatchStream and log a message or something.

amol-: I think that might be its own recipe, like "How to stream data generated on the fly". At that point it  could also be a smaller example with just `do_get`

lidavidm: That makes sense. I've removed that part of the example.

amol-:
What's the state of this recipe? Does it need further review or work?

lidavidm:
Sorry for the delay - I did update it based on your feedback, so I think it's ready for another look.

PR #123, opened 2022-01-17
[Python][Flight] Add an 'HTTP basic auth' example

### LINKED ISSUE #121: [Python][Flight] Add HTTP basic auth example
See https://stackoverflow.com/q/70740898/262727 where the proliferation of auth methods in Python is causing some confusion.

lidavidm:
Fixes #121.

### DIFF
--- a/python/source/flight.rst
+++ b/python/source/flight.rst
@@ -390,13 +390,13 @@ Authentication with user/password
 =================================
 
 Often, services need a way to authenticate the user and identify who
-they are. Flight provides several ways to implement authentication;
-the simplest uses a user-password scheme. At startup, the client
-authenticates itself with the server using a username and
-password. The server returns an authorization token to include on
-future requests.
+they are. Flight provides :doc:`several ways to implement
+authentication <pyarrow:format/Flight>`; the simplest uses a
+user-password scheme. At startup, the client authenticates itself with
+the server using a username and password. The server returns an
+authorization token to include on future requests.
 
-.. warning:: Authentication should only be used over an encrypted
+.. warning:: Authentication should only be used over a secure encrypted
              channel, i.e. TLS should be enabled.
 
 .. note:: While the scheme is described as "`(HTTP) basic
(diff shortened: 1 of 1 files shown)


lidavidm:
<details>
<summary>Rendered</summary>

image

</details>

python/source/flight.rst
```diff
@@ -387,3 +387,214 @@ stream as it arrives, instead of reading them all into a table:
...
+               if username not in self.creds or password != self.creds[username]:
+                   raise pa.flight.FlightUnauthenticatedError("Unknown user or invalid password")
+               # Return a bearer token to the user for future calls.
+               # Obviously, this is insecure and easily faked. This is for
+               # demonstration only.
+               bearer_token = username
+               return BasicAuthServerMiddleware(bearer_token)
+           elif values[0] == "Bearer":
+               # An actual call. Validate the bearer token returned to the user.
+               # The token is just the username.
```
pitrou, line 463: It shouldn't be difficult to generate an actual random token, actually. Here is an example:
https://gist.github.com/pitrou/65083ea74fe75ee4f67b6a3e1040ed58

pitrou (commented):
Looks great to me, just a couple suggestions.

python/source/flight.rst
lidavidm: Thanks! I updated it along those lines. (I kept the warning about it being insecure, though, since I don't want to make guarantees we can't back up.)

python/source/flight.rst
```diff
@@ -387,3 +387,214 @@ stream as it arrives, instead of reading them all into a table:
 
     # Shutdown the server
     server.shutdown()
+
+Authentication with user/password
+=================================
+
+Often, services need a way to authenticate the user and identify who
+they are. Flight provides several ways to implement authentication;
```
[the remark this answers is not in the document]
lidavidm: I linked to the Flight docs, but the docs are rather spare; I filed https://issues.apache.org/jira/browse/ARROW-15628 to improve that.

pitrou (approved):
+1, thanks for the update!

PR #124, opened 2022-01-18
[C++][Flight] Add basic Flight service

### LINKED ISSUE #122: [C++][Flight] Add Flight/C++ examples
We should add the same examples that Python/R/etc. have.

lidavidm:
Also fixes a bug(?) with merging outputs.

Fixes #122.

### DIFF
--- a/cpp/code/flight.cc
+++ b/cpp/code/flight.cc
@@ -1,0 +1,300 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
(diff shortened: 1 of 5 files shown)


lidavidm:
<details>
<summary>Rendered</summary>

image

</details>

cpp/code/flight.cc
```diff
@@ -0,0 +1,300 @@
...
+    return root_->DeleteFile(key);
+  }
+
+  std::shared_ptr<arrow::fs::FileSystem> root_;
+};  // end ParquetStorageService
+
+TEST(ParquetStorageServiceTest, PutGetDelete) {
+  StartRecipe("ParquetStorageService::StartServer");
+  auto fs = std::make_shared<arrow::fs::LocalFileSystem>();
+  ASSERT_OK(fs->CreateDir("./flight_datasets/"));
```
pitrou, line 187: Wouldn't it be better to define a helper function returning a `Status`, and use `ARROW_RETURN_NOT_OK`  / `ARROW_ASSIGN_OR_RAISE` instead of asserts?

Related, though the solution there is different: https://github.com/apache/arrow-cookbook/pull/52

pitrou (commented):
LGTM except for a suggestion.

cpp/code/flight.cc
lidavidm: Ah, yes, having that helper in #52 would clean things up here. A helper function would work fine right now as well.

lidavidm: Updated (finally, sorry for the delay).

PR #133, opened 2022-01-25
[Java] Cookbook CI

amol-:


amol-:
Will add deploy part as soon as I confirmed that the testing part works as expected

PR #136, opened 2022-01-26
[Java]: Java cookbook for create arrow read/write IPC format

davisusanibar:
Java cookbook for create arrow read/write IPC format


davisusanibar:
https://issues.apache.org/jira/browse/ARROW-14882

java/source/io.rst
```diff
@@ -0,0 +1,578 @@
...
+
+========================
+Reading and writing data
+========================
+
+Recipes related to reading and writing data from disk using Apache Arrow.
+
+Arrow defines two types of binary formats for serializing record batches `IPC <https://arrow.apache.org/docs/java/ipc.html>`_: Streaming format / File or Random Access format
+
+Arrow vectors can be serialized to disk as the Arrow IPC format. Such files can be directly memory-mapped when read.
```
lidavidm, line 11: I think we can consolidate this section, or we can just leave it out in favor of linking to the Arrow documentation. 

Perhaps something like this:

```
The `Arrow IPC format <https://arrow.apache.org/docs/java/ipc.html>`_ defines two types of binary formats for serializing Arrow data: the streaming format and the file format (or random access format). Such files can be directly memory-mapped when read.
```

java/source/io.rst
```diff
@@ -0,0 +1,578 @@
...
+
+.. contents::
+
+Writing
+=======
+
+Both libraries writing random access file and streaming format offer the same API
```
lidavidm, line 18: "Both writing file and streaming formats use the same API."

java/source/io.rst
```diff
@@ -0,0 +1,578 @@
...
+    import java.io.File;
+    import java.io.FileNotFoundException;
+    import java.io.FileOutputStream;
+    import java.io.IOException;
+
+    // Create and populate data:
+    Field name = new Field("name", FieldType.nullable(new ArrowType.Utf8()), null);
+    Field age = new Field("age", FieldType.nullable(new ArrowType.Int(32, true)), null);
+    Schema schemaPerson = new Schema(asList(name, age));
+    RootAllocator rootAllocator = new RootAllocator(Long.MAX_VALUE);
```
lidavidm, line 47: Hmm, should we try to always close allocators, roots, etc. to make sure we always show the "right" way to do things? It does make things a little more verbose, though.

java/source/io.rst
```diff
@@ -0,0 +1,578 @@
...
+    IntVector ageVector = (IntVector) vectorSchemaRoot.getVector("age");
+    ageVector.allocateNew(3);
+    ageVector.set(0, 10);
+    ageVector.set(1, 20);
+    ageVector.set(2, 30);
+    ageVector.setValueCount(3);
+    vectorSchemaRoot.setRowCount(3);
+    File file = new File("randon_access_to_file.arrow");
+    try (FileOutputStream fileOutputStream = new FileOutputStream(file);
+         ArrowFileWriter writer = new ArrowFileWriter(vectorSchemaRoot, null, /*WritableByteChannel out*/fileOutputStream.getChannel())
```
lidavidm, line 64: nit, but for this sort of thing, I think it would be clearer if the parameter-comment was on the `null` and not on the channel

java/source/io.rst
```diff
@@ -0,0 +1,578 @@
...
+    nameVector.set(1, "Gladis".getBytes());
+    nameVector.set(2, "Juan".getBytes());
+    nameVector.setValueCount(3);
+    IntVector ageVector = (IntVector) vectorSchemaRoot.getVector("age");
+    ageVector.allocateNew(3);
+    ageVector.set(0, 10);
+    ageVector.set(1, 20);
+    ageVector.set(2, 30);
+    ageVector.setValueCount(3);
+    vectorSchemaRoot.setRowCount(3);
```
lidavidm, line 61: nit, but `VectorSchemaRoot.setRowCount` sets the vector's row counts, if we want to be a little more terse

java/source/io.rst
```diff
@@ -0,0 +1,578 @@
...
+    File file = new File("randon_access_to_file.arrow");
+    try (FileOutputStream fileOutputStream = new FileOutputStream(file);
+         ArrowFileWriter writer = new ArrowFileWriter(vectorSchemaRoot, null, /*WritableByteChannel out*/fileOutputStream.getChannel())
+    ){
+        writer.start();
+        for (int i=0; i<10; i++){
+            writer.writeBatch();
```
lidavidm, line 68: We should clearly note (either as a comment, or in the prose) that this will write the same batch 10 times, because it implicitly writes the current contents of the associated VectorSchemaRoot as a single record batch

java/source/io.rst
```diff
@@ -0,0 +1,578 @@
...
+        }
+        writer.end();
+        System.out.println(writer.getRecordBlocks().size());
```
lidavidm, line 71:
```suggestion
        System.out.println("Record batches written: " + writer.getRecordBlocks().size());
```

java/source/io.rst
```diff
@@ -0,0 +1,578 @@
...
+    vectorSchemaRoot.setRowCount(3);
+    File file = new File("randon_access_to_file.arrow");
+    try (FileOutputStream fileOutputStream = new FileOutputStream(file);
+         ArrowFileWriter writer = new ArrowFileWriter(vectorSchemaRoot, null, /*WritableByteChannel out*/fileOutputStream.getChannel())
+    ){
+        writer.start();
+        for (int i=0; i<10; i++){
+            writer.writeBatch();
+        }
+        writer.end();
```
lidavidm, line 70: Doesn't ArrowFileWriter's AutoCloseable implementation call `end`?

java/source/io.rst
```diff
@@ -0,0 +1,578 @@
...
+.. testoutput::
+
+    2936
+
+Reading
+=======
+
+Both libraries writing random access file and streaming format offer the same API
+with the difference that random acces file offer access to the entire payload and
+we could move where do we need to go
```
lidavidm, line 270:
```suggestion
Reading the random access format and streaming format both offer the same API,
with the difference that random access files also offer access to any record batch by index.
```

java/source/io.rst
```diff
@@ -0,0 +1,578 @@
...
+    import static java.util.Arrays.asList;
+    import org.apache.arrow.vector.ipc.ArrowFileWriter;
+    import java.io.File;
+    import java.io.FileInputStream;
+    import java.io.FileNotFoundException;
+    import java.io.FileOutputStream;
+    import java.io.IOException;
+    import org.apache.arrow.vector.ipc.ArrowFileReader;
+
+    // Read data
```
lidavidm, line 298: nit: the comment here is misleading/doesn't offer much clarity

java/source/io.rst
```diff
@@ -0,0 +1,578 @@
...
+            // Generate data or modify the root or use a VectorLoader to get fresh data from somewhere else
+            writer.writeBatch();
+        }
+        writer.end();
+
+        // read
+        try (FileInputStream fileInputStreamForStream = new FileInputStream(file);
+             ArrowStreamReader reader = new ArrowStreamReader(fileInputStreamForStream, rootAllocator)){
+            // read the batch
+            reader.loadNextBatch();
```
lidavidm, line 490: Can we demonstrate iterating through all batches in the file?

java/source/io.rst
davisusanibar: Changed

java/source/io.rst
davisusanibar: Changed

java/source/io.rst
davisusanibar: Changed

java/source/io.rst
davisusanibar: Deleted

java/source/io.rst
davisusanibar: Deleted

java/source/io.rst
davisusanibar: Thanks, changed

java/source/io.rst
davisusanibar: Added

java/source/io.rst
```diff
@@ -0,0 +1,610 @@
...
+        System.out.println(writer.getRecordBlocks().size());
+    } catch (FileNotFoundException e) {
+        e.printStackTrace();
+    } catch (IOException e) {
+        e.printStackTrace();
+    }
+
+.. testoutput::
+
+    10
```
lidavidm, line 133: "Record batches written: 10"?

lidavidm: (Ditto for below examples.)

java/source/io.rst
lidavidm: I don't see the change here. It should be something like `while(reader.loadNextBatch()) { // ... }`, right?

java/source/io.rst
davisusanibar: Changed

java/source/io.rst
```diff
@@ -0,0 +1,610 @@
...
+    nameVector.set(1, "Gladis".getBytes());
+    nameVector.set(2, "Juan".getBytes());
+    IntVector ageVector = (IntVector) vectorSchemaRoot.getVector("age");
+    ageVector.allocateNew(3);
+    ageVector.set(0, 10);
+    ageVector.set(1, 20);
+    ageVector.set(2, 30);
+    vectorSchemaRoot.setRowCount(3);
+    File file = new File("randon_access_to_file.arrow");
+    try (FileOutputStream fileOutputStream = new FileOutputStream(file);
```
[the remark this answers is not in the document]
davisusanibar: We need to create the file to then validate that, another option could be to upload a arrow file to java resources and only read that, but in this manner we are not depending of another resources.

java/source/io.rst
```diff
@@ -0,0 +1,610 @@
...
+            writer.writeBatch();
+        }
+        writer.end();
+
+        // read
+        try (FileInputStream fileInputStream = new FileInputStream(file);
+             ArrowFileReader reader = new ArrowFileReader(fileInputStream.getChannel(), rootAllocator)
+        ){
+            // read the 2nd batch or the index that you need according to you write
+            ArrowBlock block = reader.getRecordBlocks().get(0);
```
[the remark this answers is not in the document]
davisusanibar: Updated

java/source/io.rst
davisusanibar: Updated

java/source/io.rst
```diff
@@ -298,14 +303,56 @@ Read - From File
...
+                ArrowRecordBatch arrowRecordBatch2 = vectorUnloader2.getRecordBatch();
+                VectorLoader vectorLoader2 = new VectorLoader(vectorSchemaRoot);
+                vectorLoader2.load(arrowRecordBatch2);
+            }
+            if (i==2){
+                VectorUnloader vectorUnloader3 = new VectorUnloader(vectorSchemaRoot3);
+                ArrowRecordBatch arrowRecordBatch3 = vectorUnloader3.getRecordBatch();
+                VectorLoader vectorLoader3 = new VectorLoader(vectorSchemaRoot);
+                vectorLoader3.load(arrowRecordBatch3);
+            }
```
lidavidm, line 355: At this point, just don't use the loop.

java/source/io.rst
```diff
@@ -298,14 +303,56 @@ Read - From File
...
     ){
         // write
         writer.start();
-        for (int i=0; i<10; i++){
+        for (int i=0; i<3; i++){
             // Generate data or modify the root or use a VectorLoader to get fresh data from somewhere else
+            if (i==1){
+                VectorUnloader vectorUnloader2 = new VectorUnloader(vectorSchemaRoot2);
+                ArrowRecordBatch arrowRecordBatch2 = vectorUnloader2.getRecordBatch();
+                VectorLoader vectorLoader2 = new VectorLoader(vectorSchemaRoot);
```
lidavidm, line 347: FWIW, why not clean things up and unload the batch when we first create the other roots?

lidavidm: Though, a suggestion: why not just `ageVector.set(0, i * 10); ageVector.set(1, i * 10 + 1); ...` etc in the loop and demonstrate modifying the root?

lidavidm: We could have a separate recipe to demonstrate VectorLoader/VectorUnloader.

java/source/io.rst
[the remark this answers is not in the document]
lidavidm: I'm suggesting this since right now the recipes are combining read and write. I know that recipes should be self-contained, but the recipe is much less helpful if the reader has to scan over it carefully to figure out which part of the recipe is relevant and which part is just noise.

java/source/io.rst
```diff
@@ -457,18 +504,19 @@ Read - From File
...
 
         // read
         try (FileInputStream fileInputStreamForStream = new FileInputStream(file);
              ArrowStreamReader reader = new ArrowStreamReader(fileInputStreamForStream, rootAllocator)){
-            // read the batch
-            reader.loadNextBatch();
-            VectorSchemaRoot vectorSchemaRootRecover = reader.getVectorSchemaRoot();
-            System.out.print(vectorSchemaRootRecover.contentToTSVString());
+            while(reader.loadNextBatch()){
+                // read the batch (on the next example you could see how to use a VectorLoader to get fresh data)
```
lidavidm, line 516: I don't think this is relevant to this recipe? It really sounds like we should have more recipes for loader/unloader.

java/source/io.rst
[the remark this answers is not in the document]
davisusanibar: Added a arrow file to resources to only consider read recipe on the code

java/source/io.rst
davisusanibar: Updated

java/source/io.rst
davisusanibar: Added suggestion and also added a separate recipe to demonstrate VectorLoader/VectorUnloader.

java/source/io.rst
davisusanibar: Deleted and added a separate recipe to demonstrate VectorLoader/VectorUnloader.

java/source/io.rst
```diff
@@ -0,0 +1,646 @@
...
+            ageVector.set(1, 20);
+            ageVector.set(2, 30);
+            vectorSchemaRoot.setRowCount(3);
+            File file = new File("randon_access_to_file.arrow");
+            try (FileOutputStream fileOutputStream = new FileOutputStream(file);
+                 ArrowFileWriter writer = new ArrowFileWriter(vectorSchemaRoot, null, fileOutputStream.getChannel())
+            ) {
+                writer.start();
+                for (int i = 0; i < 10; i++) {
+                    // Generate data or modify the root or use a VectorLoader to get fresh data from somewhere else
```
lidavidm, line 64: Can we instead make it explicit that this will write the exact same data 10x, and that the data comes from the root?

lidavidm: The concept of 'the root is mutable' is best introduced with separate examples

java/source/io.rst
```diff
@@ -0,0 +1,646 @@
...
+                    writer.writeBatch();
+                }
+                writer.end();
+                System.out.println("Record batches written: " + writer.getRecordBlocks().size());
+            } catch (FileNotFoundException e) {
+                e.printStackTrace();
+            } catch (IOException e) {
```
lidavidm, line 71: I think FileNotFoundException subclasses IOException? Can we consolidate the catches?

java/source/io.rst
```diff
@@ -0,0 +1,646 @@
...
+    import java.io.FileNotFoundException;
+    import java.io.FileOutputStream;
+    import java.io.IOException;
+
+    try(RootAllocator rootAllocator = new RootAllocator(Long.MAX_VALUE)){
+        File file = new File("./thirdpartydeps/arrowfiles/data1.arrow");
+        try (FileInputStream fileInputStream = new FileInputStream(file);
+             ArrowFileReader reader = new ArrowFileReader(fileInputStream.getChannel(), rootAllocator)
+        ){
+            System.out.println("Record batches readed: " + reader.getRecordBlocks().size());
```
lidavidm, line 289:
```suggestion
            System.out.println("Record batches in file: " + reader.getRecordBlocks().size());
```

lidavidm: and ditto below

java/source/io.rst
```diff
@@ -0,0 +1,646 @@
...
+    import org.apache.arrow.vector.VectorSchemaRoot;
+    import org.apache.arrow.vector.util.ByteArrayReadableSeekableByteChannel;
+
+    import java.io.IOException;
+    import java.nio.file.Files;
+    import java.nio.file.Path;
+    import java.nio.file.Paths;
+
+    try(RootAllocator rootAllocator = new RootAllocator(Long.MAX_VALUE)) {
+        Path path = Paths.get("./thirdpartydeps/arrowfiles/data1.arrow");
```
lidavidm, line 336: nit, but maybe name these after the example they're used in at least? or name them after the dataset? (See the ones in the C++ subdirectory, for instance - and maybe we could borrow those?)

java/source/io.rst
```diff
@@ -0,0 +1,646 @@
...
+    import java.io.ByteArrayInputStream;
+    import java.io.IOException;
+    import java.nio.file.Files;
+    import java.nio.file.Path;
+    import java.nio.file.Paths;
+
+    try(RootAllocator rootAllocator = new RootAllocator(Long.MAX_VALUE)) {
+        Path path = Paths.get("./thirdpartydeps/arrowfiles/data2.arrow");
+        try (ArrowStreamReader reader = new ArrowStreamReader(new ByteArrayInputStream(Files.readAllBytes(path)), rootAllocator)){
+            while(reader.loadNextBatch()){
```
lidavidm, line 425: more of a nit, but it would be nice to have some way to autoformat the code in these examples

java/source/io.rst
```diff
@@ -0,0 +1,646 @@
...
+    Mara    15
+    name    age
+    Raul    34
+    Jhon    29
+    Thomy    33
+
+Reading Parquet File
+********************
+
+Please check :ref:`arrow-dataset`
```
lidavidm, line 451:
```
:doc:`./dataset`
```
might work better

java/source/io.rst
```diff
@@ -0,0 +1,646 @@
...
+
+Writing Fresh Data
```
lidavidm, line 453: These two examples aren't really part of IO per se. This is about the basic usage of the Arrow Java library: roots and vectors are mutable.

lidavidm: I'd rather have examples in the basic section that show how to use VectorLoader/VectorUnloader, and how to modify the data in the root (I suppose we have that one already). If we have to show a combined example, I'd rather modify the existing write examples (e.g. they could fill a vector with random integers on each iteration).

java/source/io.rst
davisusanibar: Added text: All the data that is going to write comes from the root and as we were not changing root data we are writing the same data 10x times

java/source/io.rst
davisusanibar: Updated

java/source/io.rst
davisusanibar: Added text when are going to use that: We are providing a path with auto generated arrow files for testing purposes, change that at your convenience.

java/source/io.rst
davisusanibar: I was thinking about that, ticket created https://issues.apache.org/jira/browse/ARROW-15713

java/source/io.rst
davisusanibar: Changed

java/source/io.rst
davisusanibar: We are going to add examples for VectorLoader/VectorUnloader, and how to modify the data in the root.

What about if maintain this as it is, because these could be help something when need to persist fresh data thru io files.

java/source/io.rst
lidavidm: Ok, but it seems data1 is a random access file and data2 is a streaming file? Can we rename them to reflect that?

java/source/io.rst
lidavidm: To me, the recipe feels very unnatural. For instance, why not just have this in the main recipe up above? Then we don't need two separate recipes.

```
                // Delete the set() calls above
                writer.start();
                for (int i = 0; i < 3; i++) {
                    nameVector.set(0, ("New-"+(i+1)).getBytes());
                    nameVector.set(1, ("New-"+(i+1)).getBytes());
                    nameVector.set(2, ("New-"+(i+1)).getBytes());
                    ageVector.set(0, (i+2)*10);
                    ageVector.set(1, (i+2)*20);
                    ageVector.set(2, (i+2)*30);
                    writer.writeBatch();
                }
```

java/source/io.rst
```diff
@@ -0,0 +1,630 @@
...
+            nameVector.allocateNew(3);
+            nameVector.set(0, "Raul".getBytes());
+            nameVector.set(1, "Johao".getBytes());
+            nameVector.set(2, "Rafael".getBytes());
+            IntVector ageVector = (IntVector) vectorSchemaRoot.getVector("age");
+            ageVector.allocateNew(3);
+            ageVector.set(0, 10);
+            ageVector.set(1, 20);
+            ageVector.set(2, 30);
+            vectorSchemaRoot.setRowCount(3);
```
lidavidm, line 570: This data is never used right? Nor is the schema above?

java/source/io.rst
lidavidm: The last example we could keep, but we should create the writer after the reader and use the reader's schema to create the writer's root.

davisusanibar: Yes, but in the last case the data flow is:

- First the user have partial data to write
- Second the user obtain more data from a external file a add this data to the initial data

Please let me know if this is ok

lidavidm: To me that is a very specific use case, and it should be clear from general principles. We can't cover every possible combination of things people want to do. So I think we should to break down our examples into very focused parts, so that they can be applied in more situations, rather than trying to make examples for every situation we can think of.

java/source/io.rst
davisusanibar: Changed

java/source/io.rst
davisusanibar: Yes is difficult to cover all scenarios in our cookbook. We could improve and cover more custom escenarios in next versions.

Please let me know if we could continue as it is.

java/source/io.rst
```diff
@@ -0,0 +1,542 @@
...
+            File file = new File("randon_access_to_file.arrow");
+            try (FileOutputStream fileOutputStream = new FileOutputStream(file);
+                 ArrowFileWriter writer = new ArrowFileWriter(vectorSchemaRoot, null, fileOutputStream.getChannel())
+            ) {
+                writer.start();
+                for (int i = 0; i < 10; i++) {
+                    // All the data that is  going to write comes from the root and as we were not changing root data we are writing the same data 10x times
```
amol-, line 63: This makes the recipe just more confusing. If we have 3 rows why writing 10? It just confusing.
Also the for should probably get the length from the `vectorSchemaRoot` as in the standard case you want to write the content of a table, not a random number of rows

java/source/io.rst
```diff
@@ -0,0 +1,542 @@
...
+                 ArrowFileWriter writer = new ArrowFileWriter(vectorSchemaRoot, null, Channels.newChannel(out)))
+            {
+                writer.start();
+                for (int i=0; i<10; i++){
+                    nameVector.set(0, ("New-"+(i+1)).getBytes());
+                    nameVector.set(1, ("New-"+(i+1)).getBytes());
+                    nameVector.set(2, ("New-"+(i+1)).getBytes());
+                    ageVector.set(0, (i+2)*10);
+                    ageVector.set(1, (i+2)*20);
+                    ageVector.set(2, (i+2)*30);
```
amol-, line 118: I think that changing the data _while_ writing is just making the recipe more confusing. Let's stick to how to write an existing `VectorSchemaRoot` to a buffer. See also https://github.com/apache/arrow-cookbook/pull/136/files#r809885738

java/source/io.rst
```diff
@@ -0,0 +1,542 @@
...
+            ageVector.set(0, 10);
+            ageVector.set(1, 20);
+            ageVector.set(2, 30);
+            vectorSchemaRoot.setRowCount(3);
+            File file = new File("streaming_to_file.arrow");
+            try (FileOutputStream fileOutputStream = new FileOutputStream(file);
+                 ArrowStreamWriter writer = new ArrowStreamWriter(vectorSchemaRoot, null, fileOutputStream.getChannel())
+            ){
+                writer.start();
+                for (int i=0; i<10; i++){
```
amol-, line 176: same as before

java/source/io.rst
```diff
@@ -0,0 +1,542 @@
...
+            VarCharVector nameVector = (VarCharVector) vectorSchemaRoot.getVector("name");
+            nameVector.allocateNew(3);
+            IntVector ageVector = (IntVector) vectorSchemaRoot.getVector("age");
+            ageVector.allocateNew(3);
+            vectorSchemaRoot.setRowCount(3);
+            try (ByteArrayOutputStream out = new ByteArrayOutputStream();
+                 ArrowStreamWriter writer = new ArrowStreamWriter(vectorSchemaRoot, null, Channels.newChannel(out))
+            ){
+                writer.start();
+                for (int i=0; i<10; i++){
```
amol-, line 226: here too ;)

java/source/io.rst
davisusanibar: In this case the data are containing two columns and 3 rows per columns is assigned to the root holder.

In this case with writer.writeBatch() one batch were added with 3 records. We put a for to show how another 09 batch were added with the same 3 records each, it is because that root was not changed.

Please let me know if after this comment is needed to delete the for..

java/source/io.rst
```diff
@@ -0,0 +1,542 @@
...
+    Mara    15
+    name    age
+    Raul    34
+    Jhon    29
+    Thomy    33
+
+Reading Parquet File
+********************
+
+Please check :doc:`JNI Dataset <./dataset>`
```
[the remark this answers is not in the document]
davisusanibar: Changed

java/source/io.rst
davisusanibar: Please confirm https://github.com/apache/arrow-cookbook/pull/136#discussion_r809969275 to I delete for lines of code

java/source/io.rst
amol-: I don't see the reason why we would want to show that 10 batches are written if to write the data there is a need for a single batch. I think that users will just have an hard time understanding where that `10` is coming from.

davisusanibar: Clear now. Deleted.

java/source/io.rst
davisusanibar: Deleted

java/source/io.rst
davisusanibar: Deleted

java/source/io.rst
davisusanibar: Deleted

java/source/io.rst
```diff
@@ -0,0 +1,526 @@
...
+    import java.nio.channels.Channels;
+
+    try (RootAllocator rootAllocator = new RootAllocator(Long.MAX_VALUE)) {
+        // Create and populate data:
+        Field name = new Field("name", FieldType.nullable(new ArrowType.Utf8()), null);
+        Field age = new Field("age", FieldType.nullable(new ArrowType.Int(32, true)), null);
+        Schema schemaPerson = new Schema(asList(name, age));
+        try(VectorSchemaRoot vectorSchemaRoot = VectorSchemaRoot.create(schemaPerson, rootAllocator)){
+            VarCharVector nameVector = (VarCharVector) vectorSchemaRoot.getVector("name");
+            nameVector.allocateNew(3);
```
lidavidm, line 210: Should we still set the data like we do above?

java/source/io.rst
lidavidm: I think if we removed other examples of updating and writing data we should remove this one too. We can focus on just how to update the root and use vectorloader/vectorunloader in separate recipes.

java/source/io.rst
davisusanibar: Yes, updated

lidavidm (approved):
LGTM. Thank you!

PR #137, opened 2022-01-27
[Java]: Java cookbook for create arrow flight

davisusanibar:
[Java]: Java cookbook for create arrow read/write IPC format


java/source/flight.rst
```diff
@@ -0,0 +1,1699 @@
...
+============
+
+Recipes related to leveraging Arrow Flight protocol
+
+.. contents::
+
+Simple Service with Arrow Flight
+================================
+
+We are going to create: Flight Producer and Fligh Server:
```
lidavidm, line 14:
```suggestion
We are going to create: Flight Producer and Flight Server:
```

java/source/flight.rst
```diff
@@ -0,0 +1,1699 @@
...
+
+* InMemoryStore: A FlightProducer that hosts an in memory store of Arrow buffers. Used for integration testing.
```
lidavidm, line 16: We don't need to mention the integration testing part.

java/source/flight.rst
```diff
@@ -0,0 +1,1699 @@
...
+
+* ExampleFlightServer: An Example Flight Server that provides access to the InMemoryStore.
```
lidavidm, line 18: I know the integration test has a separate server class, but IMO we don't need to create a server class. It just bloats the code and doesn't really do anything useful. Just use FlightServer.

java/source/flight.rst
```diff
@@ -0,0 +1,1699 @@
...
+            System.out.println("Calling to getStream");
+            getStream(ticket).sendTo(allocator, listener);
+        }
+
+        public Stream getStream(Ticket t) {
+            ExampleTicket example = ExampleTicket.from(t);
+            FlightDescriptor d = FlightDescriptor.path(example.getPath());
+            FlightHolder h = holders.get(d);
+            if (h == null) {
+                throw new IllegalStateException("Unknown ticket.");
```
lidavidm, line 78: We should use FlightRuntimeException here (`throw CallStatus.NOT_FOUND.withDescription("").toRuntimeException()`)

java/source/flight.rst
```diff
@@ -0,0 +1,1699 @@
...
+    import org.apache.arrow.flight.FlightProducer;
+    import org.apache.arrow.flight.FlightStream;
+    import org.apache.arrow.flight.Location;
+    import org.apache.arrow.flight.PutResult;
+    import org.apache.arrow.flight.Result;
+    import org.apache.arrow.flight.Ticket;
+    import org.apache.arrow.flight.example.ExampleTicket;
+    import org.apache.arrow.flight.example.FlightHolder;
+    import org.apache.arrow.flight.example.Stream;
+    import org.apache.arrow.flight.example.Stream.StreamCreator;
```
lidavidm, line 40: Let's not import classes from `org.apache.arrow.flight.example` - treat those as internal use only. Either we copy the classes here or (preferably) inline code where appropriate to make the example more manageable, I don't think having a ton of little utility classes really helps things here.

java/source/flight.rst
```diff
@@ -0,0 +1,1699 @@
...
+            if (h == null) {
+                throw new IllegalStateException("Unknown ticket.");
+            }
+
+            return h.getStream(example);
+        }
+
+        @Override
+        public void listFlights(CallContext context, Criteria criteria, StreamListener<FlightInfo> listener) {
+            System.out.println("Calling to listFligths");
```
lidavidm, line 86:
```suggestion
            System.out.println("Calling to listFlights");
```

java/source/flight.rst
```diff
@@ -0,0 +1,1699 @@
...
+                try (VectorSchemaRoot root = flightStream.getRoot()) {
+                    final FlightHolder h = holders.computeIfAbsent(
+                            flightStream.getDescriptor(),
+                            t -> new FlightHolder(allocator, t, flightStream.getSchema(), flightStream.getDictionaryProvider()));
+
+                    creator = h.addStream(flightStream.getSchema());
+
+                    VectorUnloader unloader = new VectorUnloader(root);
+                    while (flightStream.next()) {
+                        ackStream.onNext(PutResult.metadata(flightStream.getLatestMetadata()));
```
lidavidm, line 124: IMO we can just ignore metadata for the sake of this example

java/source/flight.rst
```diff
@@ -0,0 +1,1699 @@
...
+                }
+            }
+        }
+
+        @Override
+        public void listActions(CallContext context,
+                                StreamListener<ActionType> listener) {
+            System.out.println("Calling to listActions");
+            listener.onNext(new ActionType("get", "pull a stream. Action must be done via standard get mechanism"));
+            listener.onNext(new ActionType("put", "push a stream. Action must be done via standard put mechanism"));
```
lidavidm, line 162: I don't think we need these actions, they're not used

java/source/flight.rst
```diff
@@ -0,0 +1,1699 @@
...
+
+Creating the Client
+*******************
+
+Get List Actions
+----------------
+
+Validate lists actions available on the Flight service:
+
+.. testcode::
```
lidavidm, line 255: Hmm, we have to repeat the entire contents of the server each time? I think that will get annoying very fast.

lidavidm: We could just put all of the client calls in the first code block (with println calls to separate things) and then talk about what happens afterwards.

java/source/flight.rst
```diff
@@ -0,0 +1,1699 @@
...
+    RootAllocator rootAllocator = new RootAllocator(Long.MAX_VALUE);
+    FlightClient client = FlightClient.builder(rootAllocator, Location.forGrpcInsecure("localhost", 33333)).build();
+
+    /**
+     * Lists flight information.
+     */
+    Iterable<FlightInfo> listFlights = client.listFlights(Criteria.ALL);
+
+    listFlights.forEach(t -> System.out.println(t));
+    System.out.println("Any list flight availale at this moment");
```
lidavidm, line 735: This print seems a little confusing, maybe we can print before we print each flight and say something like "Available datasets:"?

java/source/flight.rst
```diff
@@ -0,0 +1,1699 @@
...
+     * Exchange data.
+     */
+    import org.apache.arrow.flight.FlightClient;
+    import org.apache.arrow.flight.AsyncPutListener;
+
+    /**
+     * An identifier for a particular set of data.  This can either be an opaque command that generates
+     * the data or a static "path" to the data.  This is a POJO wrapper around the protobuf message with
+     * the same name.
+     */
```
lidavidm, line 1019: Let's not copy doc comments 1:1

java/source/flight.rst
davisusanibar: Changed

java/source/flight.rst
davisusanibar: Deleted

java/source/flight.rst
davisusanibar: Just using FlightServer

java/source/flight.rst
davisusanibar: Deleted

java/source/flight.rst
davisusanibar: Deleted

java/source/flight.rst
davisusanibar: Deleted

java/source/flight.rst
lidavidm: Sorry, this is the opposite of what I meant: we should demonstrate implementing FlightProducer, but we should not use anything from `org.apache.arrow.flight.example`, and we don't need to copy how it creates an "ExampleFlightServer" class. We should just directly create "FlightServer".

java/source/flight.rst
lidavidm: Yes, all code needs to work in isolation, but it's not very helpful if every example starts with a few hundred lines of copy-pasted code since then it's hard to see what the example is trying to demonstrate. The other language cookbooks don't repeat the server definition for each example, for instance. It might be easier to show the server once and then just use normal code blocks to show the client examples afterwards.

java/source/flight.rst
lidavidm: I don't see the change here.

java/source/flight.rst
davisusanibar: I got it now, let me implement on that way

java/source/flight.rst
davisusanibar: Adding all the code at the initial documentation and continue explaining the code.

Also added a ticket to create a custom plugin to be able to run java code cookbook but only show to our user the main code: https://issues.apache.org/jira/browse/ARROW-15703

java/source/flight.rst
davisusanibar: Updated

java/source/flight.rst
```diff
@@ -0,0 +1,1061 @@
...
+        }
+    }
+
+    // Server
+    Location location = Location.forGrpcInsecure("0.0.0.0", 33333);
+    Map<FlightDescriptor, DataInMemory> dataInMemory = new HashMap<>();
+    Map<String, DataInMemory> mapPojoFlightDataInMemory = new HashMap<>();
+    List<ArrowRecordBatch> listArrowRecordBatch = new ArrayList<>();
+    try (RootAllocator allocator = new RootAllocator(Long.MAX_VALUE)){
+        FlightServer flightServer = FlightServer.builder(allocator, location, new NoOpFlightProducer(){
```
lidavidm, line 98: Instead of an anonymous class, define a subclass up above and make properties like `dataInMemory` its properties.

java/source/flight.rst
```diff
@@ -0,0 +1,1061 @@
...
+        public Long getRows() {
+            return rows;
+        }
+    }
+
+    // Server
+    Location location = Location.forGrpcInsecure("0.0.0.0", 33333);
+    Map<FlightDescriptor, DataInMemory> dataInMemory = new HashMap<>();
+    Map<String, DataInMemory> mapPojoFlightDataInMemory = new HashMap<>();
+    List<ArrowRecordBatch> listArrowRecordBatch = new ArrayList<>();
```
lidavidm, line 96: This should be a local variable in acceptPut, or else we'll accumulate more data on every call.

java/source/flight.rst
```diff
@@ -0,0 +1,1061 @@
...
+        }
+        public Long getRows() {
+            return rows;
+        }
+    }
+
+    // Server
+    Location location = Location.forGrpcInsecure("0.0.0.0", 33333);
+    Map<FlightDescriptor, DataInMemory> dataInMemory = new HashMap<>();
+    Map<String, DataInMemory> mapPojoFlightDataInMemory = new HashMap<>();
```
lidavidm, line 95: This appears unused.

java/source/flight.rst
```diff
@@ -0,0 +1,1061 @@
...
+            }
+
+            @Override
+            public void getStream(CallContext context, Ticket ticket, ServerStreamListener listener) {
+                FlightDescriptor flightDescriptor = FlightDescriptor.path(new String(ticket.getBytes(), StandardCharsets.UTF_8)); // Recover data for key configured
+                if(dataInMemory.containsKey(flightDescriptor)){
+                    VectorSchemaRoot vectorSchemaRoot = VectorSchemaRoot.create(dataInMemory.get(flightDescriptor).getSchema(), allocator);
+                    listener.start(vectorSchemaRoot);
+                    for(ArrowRecordBatch arrowRecordBatch : dataInMemory.get(flightDescriptor).getListArrowRecordBatch()){
+                        vectorSchemaRoot.allocateNew();
```
lidavidm, line 125: Hmm, do we have to allocate before loading? I thought loading a batch comes with its own allocations already.

java/source/flight.rst
```diff
@@ -0,0 +1,1061 @@
...
+                        VectorLoader loader = new VectorLoader(vectorSchemaRoot);
+                        loader.load(arrowRecordBatch.cloneWithTransfer(allocator));
```
lidavidm, line 127: Similarly - do we have to clone before loading? That'll copy data but we're not mutating anything.

java/source/flight.rst
```diff
@@ -0,0 +1,1061 @@
...
+                        listener.putNext();
+                    }
+                    vectorSchemaRoot.clear();
```
lidavidm, line 130: We don't need to clear at the end.

java/source/flight.rst
```diff
@@ -0,0 +1,1061 @@
...
+            @Override
+            public Runnable acceptPut(CallContext context, FlightStream flightStream, StreamListener<PutResult> ackStream) {
+                return () -> {
+                    long rows = 0;
+                    while (flightStream.next()) {
+                        VectorUnloader unloader = new VectorUnloader(flightStream.getRoot());
+                        try (final ArrowRecordBatch arb = unloader.getRecordBatch()) {
+                            // Retain data information
+                            listArrowRecordBatch.add(arb);
+                            rows = rows + flightStream.getRoot().getRowCount();
```
lidavidm, line 108:
```suggestion
                            rows += flightStream.getRoot().getRowCount();
```

java/source/flight.rst
```diff
@@ -0,0 +1,1061 @@
...
+                        }
+                    }
+                    long finalRows = rows;
```
lidavidm, line 111: Why the extra variable?

java/source/flight.rst
```diff
@@ -0,0 +1,1061 @@
...
+                    }
+                    vectorSchemaRoot.clear();
+                    listener.completed();
+                }
+            }
+
+            @Override
+            public void doAction(CallContext context, Action action, StreamListener<Result> listener) {
+                FlightDescriptor flightDescriptor = FlightDescriptor.path(new String(action.getBody(), StandardCharsets.UTF_8)); // For recover data for key configured
+                if(dataInMemory.containsKey(flightDescriptor)) {
```
lidavidm, line 138: We might want to use ConcurrentHashMap and show 'proper' handling of multithreading.

java/source/flight.rst
```diff
@@ -0,0 +1,1061 @@
...
+                            listener.onNext(result);
+                    }
+                    listener.onCompleted();
+                }
+            }
+
+            @Override
+            public FlightInfo getFlightInfo(CallContext context, FlightDescriptor descriptor) {
+                if(!dataInMemory.containsKey(descriptor)){
+                    throw new IllegalStateException("Unknown descriptor.");
```
lidavidm, line 152: `throw CallStatus.NOT_FOUND.withDescription("Unknown descriptor").toRuntimeException()`

java/source/flight.rst
```diff
@@ -0,0 +1,1061 @@
...
+                }
+                return new FlightInfo(
+                        dataInMemory.get(descriptor).getSchema(),
+                        descriptor,
+                        Collections.singletonList(new FlightEndpoint(new Ticket(descriptor.getPath().get(0).getBytes(StandardCharsets.UTF_8)), location)), // Configure a key to map back and forward your data using Ticket argument
+                        allocator.getAllocatedMemory(),
```
lidavidm, line 158: We can just put `-1` if the size is unknown

java/source/flight.rst
```diff
@@ -0,0 +1,1061 @@
...
+        VarCharVector varCharVector = (VarCharVector) vectorSchemaRoot.getVector("name");
+        varCharVector.allocateNew(3);
+        varCharVector.set(0, "Ronald".getBytes());
+        varCharVector.set(1, "David".getBytes());
+        varCharVector.set(2, "Francisco".getBytes());
+        varCharVector.setValueCount(3);
+        vectorSchemaRoot.setRowCount(3);
+        FlightClient.ClientStreamListener listener = flightClient.startPut(FlightDescriptor.path("profiles"), vectorSchemaRoot, new AsyncPutListener());
+        listener.putNext();
+        vectorSchemaRoot.allocateNew();
```
lidavidm, line 195: Do you have to `allocateNew` the second time around? Also, `setRowCount` calls `setValueCount` for you.

java/source/flight.rst
```diff
@@ -0,0 +1,1061 @@
...
+        varCharVector.set(0, "Manuel".getBytes());
+        varCharVector.set(1, "Felipe".getBytes());
+        varCharVector.set(2, "JJ".getBytes());
+        varCharVector.setValueCount(3);
+        vectorSchemaRoot.setRowCount(3);
+        listener.putNext();
+        vectorSchemaRoot.clear();
```
lidavidm, line 202: If `clear` is for memory management, it's clearer to use try-with-resources

java/source/flight.rst
davisusanibar: This is global variable to collect all data in memory and increase data at acceptPut and expose data with getStream globally. This list is created at start time and increase per request and expired at stop time.

java/source/flight.rst
davisusanibar: I could suggest this is needed for transferOwnership. If I use `loader.load(arrowRecordBatch)` instead of `loader.load(arrowRecordBatch.cloneWithTransfer(allocator))` I see error `already closed`

java/source/flight.rst
lidavidm: Why a global though? Isn't that what the Map is for? Why are we accumulating data here?

java/source/flight.rst
lidavidm: I think you have to increment the refcount of the batch manually

java/source/flight.rst
davisusanibar: Changed

java/source/flight.rst
davisusanibar: Added try-with-resources

java/source/flight.rst
davisusanibar: Updated

java/source/flight.rst
davisusanibar: Added

java/source/flight.rst
davisusanibar: Changed

java/source/flight.rst
davisusanibar: Deleted

java/source/flight.rst
davisusanibar: Deleted

java/source/flight.rst
davisusanibar: Updated

java/source/flight.rst
```diff
@@ -0,0 +1,1061 @@
...
+.. KIND, either express or implied.  See the License for the
+.. specific language governing permissions and limitations
+.. under the License.
+
+============
+Arrow Flight
+============
+
+This section contains a number of recipes for working with Arrow
+Flight. For moreb about Flight.
```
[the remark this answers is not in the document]
davisusanibar: Updated

java/source/flight.rst
```diff
@@ -0,0 +1,1061 @@
...
+
+.. contents::
+
+Simple VectorSchemaRoot storage service with Arrow Flight
```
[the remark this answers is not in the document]
davisusanibar: Changed

java/source/flight.rst
```diff
@@ -0,0 +1,1061 @@
...
+=========================================================
+
+We'll implement a service to transfer batches flow through VectorSchemaRoot.
```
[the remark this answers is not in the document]
davisusanibar: Changed

java/source/flight.rst
```diff
@@ -0,0 +1,1061 @@
...
+    import java.util.Collections;
+    import java.util.HashMap;
+    import java.util.Iterator;
+    import java.util.List;
+    import java.util.Map;
+
+    class DataInMemory {
+        private List<ArrowRecordBatch> listArrowRecordBatch;
+        private Schema schema;
+        private Long rows;
```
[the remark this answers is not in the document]
davisusanibar: Updated

java/source/flight.rst
davisusanibar: New CookbookProducer class created

java/source/flight.rst
davisusanibar: Changed

java/source/flight.rst
davisusanibar: Deleted

java/source/flight.rst
davisusanibar: Your are right, I confuse with map variable, changed to local

java/source/flight.rst
davisusanibar: Let me test that

davisusanibar: Reviewing [Stream example](https://github.com/apache/arrow/blob/master/java/flight/flight-core/src/main/java/org/apache/arrow/flight/example/Stream.java#L163) I see they also use cloneWithTransfer

Probably this is needed by [cloneWithTransfer](https://github.com/apache/arrow/blob/master/java/vector/src/main/java/org/apache/arrow/vector/ipc/message/ArrowRecordBatch.java#L162:L168) method description mention:

Create a new ArrowRecordBatch which has the same information as this batch but whose buffers are owned by that Allocator.

java/source/demo/pom.xml
```diff
@@ -32,14 +32,29 @@
...
         </dependency>
         <dependency>
             <groupId>org.apache.arrow</groupId>
             <artifactId>arrow-algorithm</artifactId>
             <version>${arrow.version}</version>
         </dependency>
+        <dependency>
+            <groupId>org.apache.arrow</groupId>
+            <artifactId>flight-core</artifactId>
+            <version>6.0.0</version>
```
lidavidm, line 46: 6.0.0?

java/source/flight.rst
```diff
@@ -0,0 +1,1050 @@
...
+        public long getRows() {
+            return rows;
+        }
+    }
+
+    // Server
+    Location location = Location.forGrpcInsecure("0.0.0.0", 33333);
+    Map<FlightDescriptor, DataInMemory> dataInMemory = new ConcurrentHashMap<>();
+    try (RootAllocator allocator = new RootAllocator(Long.MAX_VALUE)){
+        class CookbookProducer extends NoOpFlightProducer {
```
lidavidm, line 98: Can't we move this up alongside DataInMemory? (We can add a constructor to give it any variables needed.)

java/source/flight.rst
```diff
@@ -0,0 +1,1050 @@
...
+            return schema;
+        }
+        public long getRows() {
+            return rows;
+        }
+    }
+
+    // Server
+    Location location = Location.forGrpcInsecure("0.0.0.0", 33333);
+    Map<FlightDescriptor, DataInMemory> dataInMemory = new ConcurrentHashMap<>();
```
lidavidm, line 96: Shouldn't this be a property of the producer class?

lidavidm, line 96: The interface here should be ConcurrentMap.

java/source/flight.rst
lidavidm: I think incrementing the refcount is sufficient since we are all using the same allocator here but I guess it doesn't matter too much (this will mean we have to copy all data each time though).

java/source/flight.rst
lidavidm: Wouldn't the right way to handle this be remove and catch the error? Since another thread may remove it concurrently, we shouldn't be checking before removing.

java/source/flight.rst
```diff
@@ -0,0 +1,1050 @@
...
+                            Result result = new Result("Delete completed".getBytes(StandardCharsets.UTF_8));
+                            listener.onNext(result);
+                    }
+                    listener.onCompleted();
+                }
+            }
+
+            @Override
+            public FlightInfo getFlightInfo(CallContext context, FlightDescriptor descriptor) {
+                if(!dataInMemory.containsKey(descriptor)){
```
lidavidm, line 148: Similarly we should get the value and handle the error if it doesn't exist.

java/source/flight.rst
lidavidm: FWIW because we show the entire server up front initially I think down below, we can just use normal code blocks to show snippets of the server instead.

lidavidm: Also I think what would help (for this, and for formatting the code) is to allow this directive to accept a file, and a line range, then it will be easier to handle long examples like this.

java/source/flight.rst
davisusanibar: Reviewing ConcurrentHashMap it is thread safe. Is it needed to add that validation?

java/source/flight.rst
davisusanibar: Yes we can, but another class is needed to create, because current design CookbookProducer update DataInMemory data. Could be possible maintain as it is now?

java/source/flight.rst
davisusanibar: Added as a property

davisusanibar: Updated

java/source/flight.rst
lidavidm: Why? `dataInMemory` should be a property of this class. We shouldn't be using globals from the server.

java/source/demo/pom.xml
lidavidm: We shouldn't mix Arrow versions.

java/source/flight.rst
davisusanibar: ConcurrentMap does nothing if the key is not in the map

java/source/demo/pom.xml
davisusanibar: Ok lets install that manually for our github ci

java/source/flight.rst
lidavidm: Individual operations are thread safe. Combinations of operations are not thread safe. This is a time-of-check-vs-time-of-use-error. If two threads call this simultaneously, both will check that the key exists, and then both try to remove it, and one will get an error. Instead, they should both just assume the key exists, remove it, and swallow the error if it happens.

java/source/flight.rst
lidavidm: What do you mean?

We cannot use containsKey, it isn't so useful with ConcurrentMap. If another thread adds or removes the key in between the check, then we will get unexpected errors.

java/source/flight.rst
```diff
@@ -0,0 +1,485 @@
...
+                    dataInMemory.put(flightStream.getDescriptor(), pojoFlightDataInMemory);
+                    ackStream.onCompleted();
+                };
+            }
+
+            @Override
+            public void getStream(CallContext context, Ticket ticket, ServerStreamListener listener) {
+                FlightDescriptor flightDescriptor = FlightDescriptor.path(new String(ticket.getBytes(), StandardCharsets.UTF_8)); // Recover data for key configured
+                if(dataInMemory.containsKey(flightDescriptor)){
+                    VectorSchemaRoot vectorSchemaRoot = VectorSchemaRoot.create(dataInMemory.get(flightDescriptor).getSchema(), allocator);
```
lidavidm, line 120: basically just do 

```
var dataset = dataInMemory.get(flightDescriptor);
if (dataset == null) { // raise an error }
// use dataset...
```

otherwise you have race conditions

java/source/flight.rst
```diff
@@ -0,0 +1,485 @@
...
+
+            @Override
+            public FlightInfo getFlightInfo(CallContext context, FlightDescriptor descriptor) {
+                if(!dataInMemory.containsKey(descriptor)){
+                    throw CallStatus.NOT_FOUND.withDescription("Unknown descriptor").toRuntimeException();
+                }
+                return new FlightInfo(
+                        dataInMemory.get(descriptor).getSchema(),
+                        descriptor,
+                        Collections.singletonList(new FlightEndpoint(new Ticket(descriptor.getPath().get(0).getBytes(StandardCharsets.UTF_8)), location)), // Configure a key to map back and forward your data using Ticket argument
```
lidavidm, line 153: Can we move the FlightEndpoint onto its own line?

java/source/flight.rst
```diff
@@ -0,0 +1,485 @@
...
+                        -1,
```
lidavidm, line 154:
```suggestion
                        /*bytes=*/-1,
```

java/source/flight.rst
davisusanibar: Ticket was created. Lets join ideas for cookbook v2

java/source/flight.rst
davisusanibar: Thanks for the clarification. Updated

java/source/demo/pom.xml
davisusanibar: Gihub CI configured to install flight core 7.0.0 manually

java/source/flight.rst
davisusanibar: Ok, updated

java/source/flight.rst
davisusanibar: Added

java/source/flight.rst
davisusanibar: Moved

java/source/flight.rst
davisusanibar: Updated

java/source/flight.rst
```diff
@@ -0,0 +1,511 @@
...
+        }
+        public long getRows() {
+            return rows;
+        }
+        public NoOpFlightProducer getCookbookProducer() {
+            return cookbookProducer;
+        }
+        public void setCookbookProducer(NoOpFlightProducer cookbookProducer) {
+            this.cookbookProducer = cookbookProducer;
+        }
```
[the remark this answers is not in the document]
davisusanibar: Deleted

java/source/flight.rst
```diff
@@ -0,0 +1,511 @@
...
+            this.listArrowRecordBatch = listArrowRecordBatch;
+            this.schema = schema;
+            this.rows = rows;
+        }
+        public DataInMemory(List<ArrowRecordBatch> listArrowRecordBatch, Schema schema, long rows, NoOpFlightProducer cookbookProducer) {
+            this.listArrowRecordBatch = listArrowRecordBatch;
+            this.schema = schema;
+            this.rows = rows;
+            this.cookbookProducer = cookbookProducer;
+        }
```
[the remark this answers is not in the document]
davisusanibar: Deleted

java/source/flight.rst
```diff
@@ -0,0 +1,511 @@
...
+    class CookbookProducer extends NoOpFlightProducer {
+        private RootAllocator allocator;
+        private Location location;
+
+        public CookbookProducer(RootAllocator allocator, Location location) {
+            this.allocator = allocator;
+            this.location = location;
+        }
+
+        ConcurrentHashMap<FlightDescriptor, DataInMemory> dataInMemory = new ConcurrentHashMap<>();
```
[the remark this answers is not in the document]
davisusanibar: Updated

java/source/flight.rst
```diff
@@ -0,0 +1,511 @@
...
+            return () -> {
+                long rows = 0;
+                while (flightStream.next()) {
+                    VectorUnloader unloader = new VectorUnloader(flightStream.getRoot());
+                    try (final ArrowRecordBatch arb = unloader.getRecordBatch()) {
+                        listArrowRecordBatch.add(arb);
+                        rows += flightStream.getRoot().getRowCount();
+                    }
+                }
+                DataInMemory pojoFlightDataInMemory = new DataInMemory(listArrowRecordBatch, flightStream.getSchema(), rows);
```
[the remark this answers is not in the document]
davisusanibar: Thanks, changed

java/source/flight.rst
```diff
@@ -0,0 +1,511 @@
...
+                        Result result = new Result("Delete not completed. Reason: Key did not exist.".getBytes(StandardCharsets.UTF_8));
+                        listener.onNext(result);
+                    }
+                    listener.onCompleted();
+            }
+        }
+
+        @Override
+        public FlightInfo getFlightInfo(CallContext context, FlightDescriptor descriptor) {
+            FlightEndpoint flightEndpoint = new FlightEndpoint(new Ticket(descriptor.getPath().get(0).getBytes(StandardCharsets.UTF_8)), location);
```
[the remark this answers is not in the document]
davisusanibar: Updated

java/source/flight.rst
```diff
@@ -0,0 +1,511 @@
...
+            );
+        }
+
+        @Override
+        public void listFlights(CallContext context, Criteria criteria, StreamListener<FlightInfo> listener) {
+            dataInMemory.forEach((k, v) -> {
+                        FlightInfo flightInfo = getFlightInfo(null, k);
+                        listener.onNext(flightInfo);
+                    }
+            );
```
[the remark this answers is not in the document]
davisusanibar: Added

java/source/flight.rst
```diff
@@ -0,0 +1,511 @@
...
+    Location location = Location.forGrpcInsecure("0.0.0.0", 33333);
+    try (RootAllocator allocator = new RootAllocator(Long.MAX_VALUE)){
+        FlightServer flightServer = FlightServer.builder(allocator, location, new CookbookProducer(allocator, location)).build();
+        try {
+            flightServer.start();
+            System.out.println("S1: Server (Location): Listening on port " + flightServer.getPort());
+        } catch (IOException e) {
+            e.printStackTrace();
+        }
+    }
```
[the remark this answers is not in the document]
davisusanibar: Updated

java/source/flight.rst
```diff
@@ -0,0 +1,534 @@
...
+    import java.io.IOException;
+    import java.nio.charset.StandardCharsets;
+    import java.util.ArrayList;
+    import java.util.Arrays;
+    import java.util.Collections;
+    import java.util.Iterator;
+    import java.util.List;
+    import java.util.concurrent.ConcurrentHashMap;
+
+    class Datasets {
```
lidavidm, line 73: nit: this should be a singular `Dataset`

java/source/flight.rst
```diff
@@ -0,0 +1,534 @@
...
+        private List<ArrowRecordBatch> batches;
+        private Schema schema;
+        private long rows;
```
lidavidm, line 76: nit, but we can make class fields `final` in general (ditto below)

java/source/flight.rst
```diff
@@ -0,0 +1,534 @@
...
+    class CookbookProducer extends NoOpFlightProducer {
+        private RootAllocator allocator;
+        private Location location;
+
+        public CookbookProducer(RootAllocator allocator, Location location) {
+            this.allocator = allocator;
+            this.location = location;
+        }
+
+        ConcurrentHashMap<FlightDescriptor, Datasets> datasets = new ConcurrentHashMap<>();
```
lidavidm, line 101: Initialize this in the constructor along with all the other fields, and move this field to be in the same block as all the other fields.

java/source/flight.rst
```diff
@@ -0,0 +1,534 @@
...
+        @Override
+        public Runnable acceptPut(CallContext context, FlightStream flightStream, StreamListener<PutResult> ackStream) {
+            List<ArrowRecordBatch> batches = new ArrayList<>();
+            return () -> {
+                long rows = 0;
+                while (flightStream.next()) {
+                    VectorUnloader unloader = new VectorUnloader(flightStream.getRoot());
```
lidavidm, line 108: nit, but we should only need to create one unloader (create it outside the loop)

java/source/flight.rst
```diff
@@ -0,0 +1,534 @@
...
+                ackStream.onCompleted();
+            };
+        }
+
+        @Override
+        public void getStream(CallContext context, Ticket ticket, ServerStreamListener listener) {
+            FlightDescriptor flightDescriptor = FlightDescriptor.path(
+                    new String(ticket.getBytes(), StandardCharsets.UTF_8));
+            Datasets datasets = this.datasets.get(flightDescriptor);
+            if (datasets == null) {
```
lidavidm, line 125: same here, just name it 'dataset'

java/source/flight.rst
```diff
@@ -0,0 +1,534 @@
...
+        }
+
+        @Override
+        public void listFlights(CallContext context, Criteria criteria, StreamListener<FlightInfo> listener) {
+            datasets.forEach((k, v) -> { listener.onNext(getFlightInfo(null, k)); });
+            listener.onCompleted();
+        }
+    }
+    Location location = Location.forGrpcInsecure("0.0.0.0", 33333);
+    Schema schema = new Schema(Arrays.asList(
```
lidavidm, line 178: nit, but create the schema closer to where it's actually used

java/source/flight.rst
```diff
@@ -0,0 +1,534 @@
...
+            new Field("name", FieldType.nullable(new ArrowType.Utf8()), null)));
+    try (RootAllocator allocator = new RootAllocator(Long.MAX_VALUE)){
+        // Server
+        try(FlightServer flightServer = FlightServer.builder(allocator, location,
+                new CookbookProducer(allocator, location)).build()) {
+            try {
+                flightServer.start();
+                System.out.println("S1: Server (Location): Listening on port " + flightServer.getPort());
+            } catch (IOException e) {
+                e.printStackTrace();
```
lidavidm, line 188: nit: System.exit here?

lidavidm (commented):
Thanks for the updates, let's just clean up a few final things

java/source/flight.rst
davisusanibar: Changed

java/source/flight.rst
davisusanibar: Added

java/source/flight.rst
davisusanibar: Changed

java/source/flight.rst
davisusanibar: Updated

java/source/flight.rst
davisusanibar: Changed

java/source/flight.rst
davisusanibar: Updated

java/source/flight.rst
davisusanibar: Changed

lidavidm:
Linux build is failing - do we need to `apt update` before `apt install`?

davisusanibar:
> Linux build is failing - do we need to `apt update` before `apt install`?
CI linux script updated and finished correctly linux & macos environment

lidavidm:
This doesn't fix the fundamental issue, see: https://docs.github.com/en/actions/using-github-hosted-runners/customizing-github-hosted-runners#installing-software-on-ubuntu-runners

We must apt update before apt install.

lidavidm:
Though all of the pipelines currently have that issue, I suppose we can fix that separately.

lidavidm:
Filed https://github.com/apache/arrow-cookbook/issues/162

PR #138, opened 2022-01-27
[Java]: Java cookbook for create arrow jni dataset

davisusanibar:
Java cookbook for create arrow jni

We are seeing lots of improvements on arrow java dataset jni library latest version 7.0.0.

We are able to run jni dataset 7.0.0 version on latest github ci images without problems and not needed to install or upgrade custom dependencies.

We are adding github ci to support java cookbook testing for: Linux & MacOS

-------

History messages:

Proposal: Use macos-latest for jni cookbooks that offer more support for libarrow_dataset_jni library

Consider:

- Current limitation for test this libarrow_dataset_jni using github ci:

Linux: Is needed to use Ubuntu 21 for libre2.so.9 and libc.so.6, latest version at github is Ubuntu 20

OSX: Is needed to install libprotobuf.28.dylib and it is not available at brew formula, needed to create custom protobuf formula

- Library dependencies:

(base) ➜  /tmp objdump -p libarrow_dataset_jni.so | grep NEEDED
  NEEDED       liblz4.so.1
  NEEDED       libsnappy.so.1
  NEEDED       libz.so.1
  NEEDED       libzstd.so.1
  NEEDED       libutf8proc.so.2
  NEEDED       libre2.so.9
  NEEDED       libthrift-0.13.0.so
  NEEDED       libstdc++.so.6
  NEEDED       libm.so.6
  NEEDED       libgcc_s.so.1
  NEEDED       libpthread.so.0
  NEEDED       libc.so.6
  NEEDED       ld-linux-x86-64.so.2

(base) ➜  /tmp otool -L libarrow_dataset_jni.dylib
libarrow_dataset_jni.dylib:
    @rpath/libarrow_dataset_jni.600.dylib (compatibility version 600.0.0, current version 600.0.0)
    /usr/local/opt/lz4/lib/liblz4.1.dylib (compatibility version 1.0.0, current version 1.9.3)
    /usr/local/opt/snappy/lib/libsnappy.1.dylib (compatibility version 1.0.0, current version 1.1.9)
    /usr/lib/libz.1.dylib (compatibility version 1.0.0, current version 1.2.11)
    /usr/local/opt/zstd/lib/libzstd.1.dylib (compatibility version 1.0.0, current version 1.5.0)
    /usr/local/opt/protobuf/lib/libprotobuf.28.dylib (compatibility version 29.0.0, current version 29.3.0)
    /usr/local/opt/utf8proc/lib/libutf8proc.2.dylib (compatibility version 2.0.0, current version 2.4.1)
    /usr/local/opt/re2/lib/libre2.9.dylib (compatibility version 9.0.0, current version 9.0.0)
    /usr/local/opt/thrift/lib/libthrift-0.15.0.dylib (compatibility version 0.0.0, current version 0.0.0)
    /usr/local/opt/llvm/lib/libc++.1.dylib (compatibility version 1.0.0, current version 1.0.0)
    /usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1292.100.5)


lidavidm:
If there are dependency errors, could we perhaps use a Conda environment? I guess I don't know what exactly the Java binaries are built against.

java/source/jni/dataset.rst
```diff
@@ -0,0 +1,294 @@
...
+      NEEDED       libzstd.so.1
+      NEEDED       libutf8proc.so.2
+      NEEDED       libre2.so.9
+      NEEDED       libthrift-0.13.0.so
+      NEEDED       libstdc++.so.6
+      NEEDED       libm.so.6
+      NEEDED       libgcc_s.so.1
+      NEEDED       libpthread.so.0
+      NEEDED       libc.so.6
+      NEEDED       ld-linux-x86-64.so.2
```
lidavidm, line 275: To be frank, I would have expected JARs with binaries to be self-contained, so this is a little unfortunate. Maybe we need to resolve this on the packaging side?

lidavidm (commented):
Why do we need to move things into "others"?

java/source/jni/dataset.rst
davisusanibar: If we offer fat jar with all dependency included the problem could be the same if the user does not have that dependencies already configured/installed.

For example if the fat jar contains libc.so.6 and the user OS is ubuntu 20 it wont work because the  libc.so.6 needs ubuntu 21 that contains GLIBC_2.32, if not the user could obtain error message like this `./libarrow_dataset_jni.so: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by ./libarrow_dataset_jni.so)`

davisusanibar:
> If there are dependency errors, could we perhaps use a Conda environment? I guess I don't know what exactly the Java binaries are built against.

We are using images provided for github ci workflow (ubuntu 2004 & macos 11.6) and added only dependencies needed by JNI defined by yml code. If this enough for this point?

java/source/jni/dataset.rst
lidavidm: Well then, it's mostly concerning that we're building JARs that sound practically useless (especially as it is not obvious what dependencies/system requirements exactly they have), and it would be much better if we could keep the cookbook organization instead of having to split things like this. But if testing on MacOS is the only way, then so be it.

Could we just run a single pipeline on MacOS, or at least run all the non-JNI tests on both platforms so we're not doing as much work in the CI to try to split up the tests so carefully?

.github/workflows/test_javajni_cookbook.yml
```diff
@@ -0,0 +1,78 @@
...
+        run: |
+          brew install wget
+          wget https://raw.githubusercontent.com/davisusanibar/custom-recipes/main/protobuf.rb
+          HOMEBREW_NO_INSTALLED_DEPENDENTS_CHECK=false brew install -s protobuf.rb
+      - name: Validate packages 1
+        run: ls -latr /usr/local/opt/protobuf/
+      - name: Disable jni cookbooks
+        run: |
+          brew install rename
+          rename 's/\.rst/\.rstbck/' ./java/source/others/*.rst
```
lidavidm, line 72: So, do we need to do this? I don't think we need to make an `others` directory, there should be no harm in running all the tests here.

.github/workflows/test_javajni_cookbook.yml
```diff
@@ -0,0 +1,78 @@
...
+      - name: Install jni dependencies
+        run: brew install snappy utf8proc re2 thrift
+      - name: Upgrade pip
+        run: python3 -m pip install --upgrade pip
+      - name: JNI - libarrow_dataset_jni compiled with libprotobuf.28.dylib not available at brew formula
+        if: steps.cache-libprotobuf.outputs.cache-hit != 'true'
+        run: |
+          brew install wget
+          wget https://raw.githubusercontent.com/davisusanibar/custom-recipes/main/protobuf.rb
+          HOMEBREW_NO_INSTALLED_DEPENDENTS_CHECK=false brew install -s protobuf.rb
```
lidavidm, line 66: Hmm, this is either pretty brittle, or will take a long time to build Protobuf each time, no? Is there really no way we can get the right dependencies on Linux? How did we even build it in the first place then?

.github/workflows/test_javajni_cookbook.yml
davisusanibar: At the end we need a rule to mark what cookbook should be running only on macOS. For that purpose we need a mark to decide what cookbook should be run on macOS our options are: naming convention / folder convention. We implement folder convention. This is needed.

If we run this cookbook on Linux Ubuntu it fail because latest github ci images not cover dependencies needed by *.so dataset library.

.github/workflows/test_javajni_cookbook.yml
davisusanibar: Yes, it take long time to build protobuf all times (1st time it take close to 10 minutes), for this reason at the first time we upload this protobuf folder libraries to cache with `uses: actions/cache@v2`

.github/workflows/test_javajni_cookbook.yml
lidavidm: Sure, I'm just wondering - why not just run all the tests on MacOS then?

lidavidm:
> We are using images provided for github ci workflow (ubuntu 2004 & macos 11.6) and added only dependencies needed by JNI defined by yml code. If this enough for this point?

What I mean is, clearly, we built these JARs on Linux somehow, presumably with our infrastructure. If Ubuntu 20.04 doesn't have the necessary dependencies, how did we get them? For instance, does Conda contain new enough versions of dependencies?

lidavidm:
And if these JARs really are not usable without a lot of complications, frankly, I'm more inclined to drop this section for now and file a JIRA issue instead.

davisusanibar:
> What I mean is, clearly, we built these JARs on Linux somehow, presumably with our infrastructure. If Ubuntu 20.04 doesn't have the necessary dependencies, how did we get them? For instance, does Conda contain new enough versions of dependencies?

Good point. Consider that restriction that I mention is for the images provided by github ci. We could do changes on that to install/remove/upgrade libraries as needed to achieve the same as our complex ci arrow.

This small PR only consider to run jni on github macos image with some changes on latest image.

I agree that is also needed to run dataset jni on linux for that we could create a ticket and work to implement that as the same that current arrow ci implement that.

davisusanibar:
> And if these JARs really are not usable without a lot of complications, frankly, I'm more inclined to drop this section for now and file a JIRA issue instead.

Please @amol- if you could help us to move forward on this jni dataset cookbook

java/source/dataset.rst
```diff
@@ -0,0 +1,232 @@
+.. _arrow-dataset:
+
+=======
+Dataset
+=======
+
+Current java project that use JNI are:
+
+* `Arrow Java Dataset <https://arrow.apache.org/docs/dev/java/dataset.html>`_: Java implementation of Arrow Dataset API/Framework. JniLoader [arrow_cdata_jni]
+* `Arrow Java C Data Interface <https://arrow.apache.org/docs/format/CDataInterface.html>`_: Java implementation of C Data Interface. JniLoader [arrow_dataset_jni]
```
lidavidm, line 10: Sorry, what are the `JniLoader [arrow_cdata_jni]` snippets here referring to? (Also, they seem to be swapped between the two?)

Also, it's just called the `Arrow C Data Interface`. However, given this file is named "Dataset", maybe let's not even talk about the C Data Interface here - that can be a separate file.

java/source/dataset.rst
```diff
@@ -0,0 +1,232 @@
+.. _arrow-dataset:
+
+=======
+Dataset
+=======
+
+Current java project that use JNI are:
+
+* `Arrow Java Dataset <https://arrow.apache.org/docs/dev/java/dataset.html>`_: Java implementation of Arrow Dataset API/Framework. JniLoader [arrow_cdata_jni]
```
lidavidm, line 9:
```suggestion
* `Arrow Java Dataset <https://arrow.apache.org/docs/dev/java/dataset.html>`_: Java implementation of Arrow Datasets library. JniLoader [arrow_cdata_jni]
```

java/source/dataset.rst
```diff
@@ -0,0 +1,232 @@
...
+    import org.apache.arrow.dataset.source.DatasetFactory;
+    import org.apache.arrow.memory.RootAllocator;
+    import org.apache.arrow.vector.types.pojo.Schema;
+    import org.apache.arrow.util.AutoCloseables;
+
+    String uri = "file:" + System.getProperty("user.dir") + "/thirdpartydeps/parquetfiles/data1.parquet";
+    RootAllocator rootAllocator = new RootAllocator(Long.MAX_VALUE);
+    DatasetFactory datasetFactory = new FileSystemDatasetFactory(rootAllocator, NativeMemoryPool.getDefault(), FileFormat.PARQUET, uri);
+    Schema schema = datasetFactory.inspect();
+    AutoCloseables.close(datasetFactory);
```
lidavidm, line 36: We should close both the factory and the allocator here. Alternatively, we could just use try-with-resources.

java/source/dataset.rst
```diff
@@ -0,0 +1,232 @@
...
+=======
+
+Current java project that use JNI are:
+
+* `Arrow Java Dataset <https://arrow.apache.org/docs/dev/java/dataset.html>`_: Java implementation of Arrow Dataset API/Framework. JniLoader [arrow_cdata_jni]
+* `Arrow Java C Data Interface <https://arrow.apache.org/docs/format/CDataInterface.html>`_: Java implementation of C Data Interface. JniLoader [arrow_dataset_jni]
+
+.. contents::
+
+Schema
```
lidavidm, line 14: I think we should reorganize this a bit. This first part is about constructing a dataset, not about schemas per se. 

When constructing a dataset, we can pass in one or more files and a file format. A dataset needs a schema. We can explicitly ask the dataset factory to inspect the files and return a schema. We should still construct the dataset with `DatasetFactory.finish(schema).` That's the first recipe.

We can add a recipe about letting the factory infer the schema. We don't call `inspect` and just call `finish`.

Then the second recipe can become "Getting the schema of a dataset". The JNI bindings don't expose this except by starting a scan, so we demonstrate that.

lidavidm: We should also add a recipe where we construct a factory from more than one file.

java/source/dataset.rst
```diff
@@ -0,0 +1,232 @@
...
+.. testoutput::
+
+    Schema<id: Int(32, true), name: Utf8>(metadata: {parquet.avro.schema={"type":"record","name":"User","namespace":"org.apache.arrow.dataset","fields":[{"name":"id","type":["int","null"]},{"name":"name","type":["string","null"]}]}, writer.model.name=avro})
+
+Query Parquet File
+==================
+
+Let query information for a parquet file usign arrow dataset module
+
+Query Data Size
```
lidavidm, line 80: We should name this "counting rows" or something. Though, I don't think it's a useful recipe to show in the first place. Particularly because specifying batchSize = 1 will be slow.

java/source/dataset.rst
```diff
@@ -0,0 +1,232 @@
...
+            int batchRowCount = vsr.getRowCount();
+            for (int i = 0; i < fieldCount; i++) {
+                FieldVector vector = vsr.getVector(i);
+                for (int j = 0; j < batchRowCount; j++) {
+                    Object object = vector.getObject(j);
+                    System.out.println(object);
+                }
+            }
+        }
+    }
```
lidavidm, line 163: We should do this iteratively instead of collecting all data into a list.

java/source/dataset.rst
```diff
@@ -0,0 +1,232 @@
...
+    import org.apache.arrow.vector.types.pojo.Schema;
+
+    import java.util.List;
+    import java.util.stream.Collectors;
+    import java.util.stream.StreamSupport;
+
+    String uri = "file:" + System.getProperty("user.dir") + "/thirdpartydeps/parquetfiles/data1.parquet";
+    RootAllocator rootAllocator = new RootAllocator(Long.MAX_VALUE);
+    DatasetFactory datasetFactory = new FileSystemDatasetFactory(rootAllocator, NativeMemoryPool.getDefault(), FileFormat.PARQUET, uri);
+    ScanOptions options = new ScanOptions(1);
```
lidavidm, line 144: Let's not use batchSize = 1. We shouldn't demonstrate suboptimal settings to people.

java/source/jni/dataset.rst
davisusanibar: We are able to run jni dataset 7.0.0 version on latest github ci images without problems and not needed to install or upgrade custom dependencies.

.github/workflows/test_javajni_cookbook.yml
davisusanibar: Added support to run cookbooks on Linux and MacOS

java/source/dataset.rst
davisusanibar: Changed

java/source/dataset.rst
davisusanibar: Added try-with-resources

java/source/dataset.rst
davisusanibar: Added a recipe for dataset
Added a recipe for construct a factory for a directory reading

java/source/dataset.rst
davisusanibar: Deleted

java/source/dataset.rst
davisusanibar: Changed to print root data after loading

java/source/dataset.rst
```diff
@@ -0,0 +1,300 @@
+.. _arrow-dataset:
+
+=======
+Dataset
+=======
+
+* `Arrow Java Dataset <https://arrow.apache.org/docs/dev/java/dataset.html>`_: Java implementation of Arrow Datasets library.
```
lidavidm, line 7: Maybe make it explicit that it's JNI bindings, not a separate implementation?

Can we use Intersphinx to do this linking? (Example for Python: https://github.com/apache/arrow-cookbook/blob/main/python/source/flight.rst#authentication-with-userpassword)

java/source/dataset.rst
```diff
@@ -0,0 +1,300 @@
...
+
+.. contents::
+
+Dataset
+=======
+
+Let construct our dataset with auto-inferred schema.
```
lidavidm, line 14:
```suggestion
We can construct a dataset with an auto-inferred schema.
```

java/source/dataset.rst
```diff
@@ -0,0 +1,300 @@
...
+                }
+            }
+        }
+    }
+
+.. testoutput::
+
+    1
+
+Let construct our dataset with predefined schema.
```
lidavidm, line 44:
```suggestion
We can also explicitly discover the schema during dataset construction, and pass that to the factory. We could modify or replace the schema ourselves if needed here.
```

java/source/dataset.rst
```diff
@@ -0,0 +1,300 @@
...
+=======
+Dataset
+=======
+
+* `Arrow Java Dataset <https://arrow.apache.org/docs/dev/java/dataset.html>`_: Java implementation of Arrow Datasets library.
+
+.. contents::
+
+Dataset
+=======
```
lidavidm, line 12:
```suggestion
Constructing Datasets
=====================
```

java/source/dataset.rst
```diff
@@ -0,0 +1,300 @@
...
+                }
+            }
+        }
+    }
+
+.. testoutput::
+
+    1
+
+Getting the Schema of a Dataset
```
lidavidm, line 74: This section doesn't make sense to have anymore.

java/source/dataset.rst
```diff
@@ -0,0 +1,300 @@
...
+
+    import java.util.List;
+    import java.util.stream.Collectors;
+    import java.util.stream.StreamSupport;
+
+    String uri = "file:" + System.getProperty("user.dir") + "/thirdpartydeps/parquetfiles/data1.parquet";
+    try(RootAllocator rootAllocator = new RootAllocator(Long.MAX_VALUE);
+        DatasetFactory datasetFactory = new FileSystemDatasetFactory(rootAllocator, NativeMemoryPool.getDefault(), FileFormat.PARQUET, uri);
+        Dataset dataset = datasetFactory.finish()){
+        ScanOptions options = new ScanOptions(100);
```
lidavidm, line 170: In all these recipes, can we provide comments for parameters like this?

lidavidm:
```suggestion
        ScanOptions options = new ScanOptions(/*batchSize*/ 100);
```

lidavidm: Also, something like 1024 might be more reasonable?

java/source/dataset.rst
```diff
@@ -0,0 +1,300 @@
...
+        try(Scanner scanner = dataset.newScan(options)){
+            Schema schema = scanner.schema();
+            List<ArrowRecordBatch> batches = StreamSupport.stream(scanner.scan().spliterator(), false).flatMap(t -> Streams.stream(t.execute())).collect(Collectors.toList());
```
lidavidm, line 173: Can we iterate instead of collecting all data up front?

java/source/dataset.rst
```diff
@@ -0,0 +1,300 @@
...
+            try (VectorSchemaRoot vsr = VectorSchemaRoot.create(schema, rootAllocator)) {
```
lidavidm, line 174: Why not create the root in the same try-with-resources block as the scanner itself, and save an indentation level?

java/source/dataset.rst
```diff
@@ -0,0 +1,300 @@
...
+    Batch Size: 5
+    Batch: 1, RowCount: 3
+    Batch: 2, RowCount: 3
+    Batch: 3, RowCount: 100
+    Batch: 4, RowCount: 100
+    Batch: 5, RowCount: 50
+
+Query Data Content with Projection
+**********************************
+
```
lidavidm, line 251: We should add a short explanation here that we can choose which columns to read from the dataset.

java/source/dataset.rst
davisusanibar: Could be ok to maintain this if the user need to print wat is the schema readed by dataset library ?

java/source/dataset.rst
davisusanibar: Is there some reason for that? I saw the unit test and it was implemented in this way.

java/source/dataset.rst
davisusanibar: Changed

java/source/dataset.rst
lidavidm: I suppose. Could the recipes then be named something like "Getting Schema During Dataset Construction" and "Getting Schema from a Dataset"?

java/source/dataset.rst
lidavidm: Unit tests aren't necessarily examples of good code, and collecting all data up front means we have to have enough memory for all the data.

java/source/dataset.rst
lidavidm: I didn't see a change here, did I miss it?

davisusanibar: Sorry, were problems in pull/merge, added finally

java/source/dataset.rst
davisusanibar: Changed

java/source/dataset.rst
```diff
@@ -0,0 +1,297 @@
...
+    import java.util.stream.StreamSupport;
+
+    String uri = "file:" + System.getProperty("user.dir") + "/thirdpartydeps/parquetfiles/";
+    try(RootAllocator rootAllocator = new RootAllocator(Long.MAX_VALUE);
+        DatasetFactory datasetFactory = new FileSystemDatasetFactory(rootAllocator, NativeMemoryPool.getDefault(), FileFormat.PARQUET, uri);
+        Dataset dataset = datasetFactory.finish()){
+        ScanOptions options = new ScanOptions(/*batchSize*/ 100);
+        try(Scanner scanner = dataset.newScan(options);
+            VectorSchemaRoot vsr = VectorSchemaRoot.create(scanner.schema(), rootAllocator)){
+            List<ArrowRecordBatch> batches = StreamSupport.stream(scanner.scan().spliterator(), false).flatMap(t -> Streams.stream(t.execute())).collect(Collectors.toList());
```
lidavidm, line 220: Again, can we iterate over this directly instead of collecting to a list first? It's verbose and demonstrates bad habits (this will load all data into memory first).

lidavidm: We don't need to print the number of batches below.

java/source/dataset.rst
```diff
@@ -0,0 +1,297 @@
...
+
+    String uri = "file:" + System.getProperty("user.dir") + "/thirdpartydeps/parquetfiles/data1.parquet";
+    try(RootAllocator rootAllocator = new RootAllocator(Long.MAX_VALUE);
+        DatasetFactory datasetFactory = new FileSystemDatasetFactory(rootAllocator, NativeMemoryPool.getDefault(), FileFormat.PARQUET, uri);
+        Dataset dataset = datasetFactory.finish()){
+        String[] projection = new String[] {"name"};
+        ScanOptions options = new ScanOptions(/*batchSize*/ 100, Optional.of(projection));
+        try(Scanner scanner = dataset.newScan(options)){
+            Schema schema = scanner.schema();
+            List<ArrowRecordBatch> batches = StreamSupport.stream(scanner.scan().spliterator(), false).flatMap(t -> Streams.stream(t.execute())).collect(Collectors.toList());
```
lidavidm, line 277: Let's fix this one too.

java/source/dataset.rst
```diff
@@ -0,0 +1,297 @@
...
+    import java.util.stream.Stream;
+
+    String uri = "file:" + System.getProperty("user.dir") + "/thirdpartydeps/parquetfiles/data1.parquet";
+    try(RootAllocator rootAllocator = new RootAllocator(Long.MAX_VALUE);
+        DatasetFactory datasetFactory = new FileSystemDatasetFactory(rootAllocator, NativeMemoryPool.getDefault(), FileFormat.PARQUET, uri);
+        Dataset dataset = datasetFactory.finish()){
+        ScanOptions options = new ScanOptions(/*batchSize*/ 100);
+        try(Scanner scanner = dataset.newScan(options);
+            VectorSchemaRoot vsr = VectorSchemaRoot.create(scanner.schema(), rootAllocator)){
+            scanner.scan().forEach(scanTask-> {
```
lidavidm, line 167: Thanks for updating this.

Though, these bindings use an outdated API…we need to fix that (not here).

java/source/dataset.rst
davisusanibar: Updated

java/source/dataset.rst
davisusanibar: Updated and delete print number batches

java/source/dataset.rst
lidavidm: Filed https://issues.apache.org/jira/browse/ARROW-15745

lidavidm (approved):

java/source/dataset.rst
```diff
@@ -0,0 +1,297 @@
...
+
+    String uri = "file:" + System.getProperty("user.dir") + "/thirdpartydeps/parquetfiles/";
+    try(RootAllocator rootAllocator = new RootAllocator(Long.MAX_VALUE);
+        DatasetFactory datasetFactory = new FileSystemDatasetFactory(rootAllocator, NativeMemoryPool.getDefault(), FileFormat.PARQUET, uri);
+        Dataset dataset = datasetFactory.finish()){
+        ScanOptions options = new ScanOptions(/*batchSize*/ 100);
+        try(Scanner scanner = dataset.newScan(options);
+            VectorSchemaRoot vsr = VectorSchemaRoot.create(scanner.schema(), rootAllocator)){
+            scanner.scan().forEach(scanTask-> {
+                Stream<ArrowRecordBatch> stream = Streams.stream(scanTask.execute());
```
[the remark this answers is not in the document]
davisusanibar: Updated

lidavidm (approved):

PR #144, opened 2022-02-07
[C++] Drop gtest to 1.10 / Raise Arrow to 6.0.1

westonpace:
I had to add UseAsync back in, as its removal was based on 7.0.0, which is a touch premature as 7.0.0 is not in conda yet.  It was ok to be removed in 5.0.0 but in 6.0.1 we don't support non-async dataset writes and so the dataset write example was failing.


westonpace:
One bonus is that we can remove the download & build gtest step from the C++ build since conda has gtest 1.10 binaries.

pitrou:
Great, thank you!

PR #153, opened 2022-02-23
[C++] gRPC settings and custom endpoint examples

### LINKED ISSUE #106: [C++] Add annotated gRPC + Flight service example
Add an example of a Flight and gRPC service coexisting on the same port. Document the caveats (namely: everything should link gRPC dynamically, not statically). Talk about why this is useful (including being able to customize low-level server options).

This can be done in Java as well, but not Python. We should also document setting gRPC client options (this can be done in Java, Python, and C++). 

The caveats should be noted in the C++ docs as well (see [ARROW-14662](https://issues.apache.org/jira/browse/ARROW-14662)).

See https://github.com/apache/arrow/issues/11657 where a user ran into this.

wjones127:
Fixes #106 

This was largely based on https://github.com/apache/arrow/blob/master/cpp/examples/arrow/flight_grpc_example.cc and https://github.com/lidavidm/arrow/blob/3777e2a6662e9530b9c91cadac3f43afae2eeb28/cpp/src/arrow/flight/flight_test.cc#L1300-L1302


cpp/code/protos/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
+# Based on https://www.f-ax.de/dev/2020/11/08/grpc-plugin-cmake-support.html
```
lidavidm, line 1: Hmm, is there a licensing concern here…?

We could also base this on how Flight generates its protos. We can just include the Proto sources directly in our binary instead of as a library.

cpp/code/flight.cc
```diff
@@ -43,9 +47,9 @@ class ParquetStorageService : public arrow::flight::FlightServerBase {
   explicit ParquetStorageService(std::shared_ptr<arrow::fs::FileSystem> root)
       : root_(std::move(root)) {}
 
-  arrow::Status ListFlights(const arrow::flight::ServerCallContext&,
-                            const arrow::flight::Criteria*,
-                            std::unique_ptr<arrow::flight::FlightListing>* listings) override {
+  arrow::Status ListFlights(
+      const arrow::flight::ServerCallContext&, const arrow::flight::Criteria*,
+      std::unique_ptr<arrow::flight::FlightListing>* listings) override {
```
lidavidm, line 52: We should really set up clang-format properly

cpp/code/flight.cc
```diff
@@ -173,7 +177,41 @@ class ParquetStorageService : public arrow::flight::FlightServerBase {
   std::shared_ptr<arrow::fs::FileSystem> root_;
 };  // end ParquetStorageService
 
-arrow::Status TestPutGetDelete() {
+class HelloWorldServiceImpl : public HelloWorldService::Service {
+  grpc::Status SayHello([[maybe_unused]] grpc::ServerContext* ctx,
```
lidavidm, line 181: Instead of the attribute, we could also just declare it as `grpc::ServerContext*,...`

cpp/code/flight.cc
```diff
@@ -173,7 +177,41 @@ class ParquetStorageService : public arrow::flight::FlightServerBase {
...
+    const std::string& name = request->name();
+    if (name.empty()) {
+      return grpc::Status(grpc::StatusCode::INVALID_ARGUMENT, "Must provide a name!");
+    }
+    reply->set_reply("Hello, " + name);
+    return grpc::Status::OK;
+  }
+};  // end HelloWorldServiceImpl
+
+class HelloWorldClient {
```
lidavidm, line 192: IMO, we can just inline the client code into our sample instead of writing out a wrapper (I don't feel like the wrapper adds much)

cpp/code/flight.cc
```diff
@@ -173,7 +177,41 @@ class ParquetStorageService : public arrow::flight::FlightServerBase {
...
+    }
+    return response.reply();
+  }
+
+  protected:
+    std::unique_ptr<HelloWorldService::Stub> stub_;
+};  // end HelloWorldClient
+
+arrow::Status
+TestPutGetDelete() {
```
lidavidm, line 214: Actually - it seems like a different formatter got used here? clang-format wouldn't do this

cpp/source/flight.rst
```diff
@@ -83,3 +83,79 @@ Finally, we'll stop our server:
...
+To create a gRPC service, first define a service using protobuf.
+
+.. literalinclude:: ../code/protos/helloworld.proto
+   :language: protobuf
+   :linenos:
+   :start-at: syntax = "proto3";
+   :caption: Hello world protobuf specification
+
+Next, you'll need to compile that. If you are using CMake, consider the setup
+suggested by https://www.f-ax.de/dev/2020/11/08/grpc-plugin-cmake-support.html.
```
lidavidm, line 125: I'm not sure it's worth linking to a random blog (no offense to the blog!)

lidavidm (commented):
Thanks for doing this! I left some thoughts.

cpp/code/protos/CMakeLists.txt
wjones127: Oh and the repo with the example code has no license. Which sounds like it means there is def an license issue.

https://github.com/faaxm/exmpl-cmake-grpc

cpp/code/CMakeLists.txt
```diff
@@ -13,12 +13,21 @@ if (CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
...
+    protos/helloworld.proto
+)
+
 # Create test targets
 enable_testing()
 
 include(GoogleTest)
 
-function(RECIPE TARGET)
+function(RECIPE TARGET COMPILE_PROTOS)
```
lidavidm, line 30: This is probably good enough for now and maybe we can add keyword arguments eventually.

Though, would it work to just drop 
```
target_link_libraries(flight ...)
...
protobuf_generate(TARGET flight LANGUAGE ...)
...
```
below without adding a parameter?

cpp/source/flight.rst
```diff
@@ -83,3 +83,72 @@ Finally, we'll stop our server:
...
+
+.. recipe:: ../code/flight.cc TestClientOptions::Connect
+   :dedent: 2
+
+
+Flight Service with other gRPC endpoints
+========================================
+
+If you are using the gRPC backend, you can add other gRPC endpoints to the 
+flight server. While Flight clients won't recognize these endpoints, general
```
lidavidm, line 105:
```suggestion
Flight server. While Flight clients won't recognize these endpoints, general
```

cpp/code/flight.cc
```diff
@@ -291,4 +307,99 @@ arrow::Status TestPutGetDelete() {
...
+
+  // Create hello world service
+  HelloWorldServiceImpl grpc_service;
+  // Both services will be available on both ports
+  int hello_world_port;
+
+  options.builder_hook = [&](void* raw_builder) {
+    auto* builder = reinterpret_cast<grpc::ServerBuilder*>(raw_builder);
+    builder->AddListeningPort("0.0.0.0:5000", grpc::InsecureServerCredentials(),
+                              &hello_world_port);
```
lidavidm, line 370: Hmm, maybe we cut the extra port from this recipe (and slim it down a bit)? The important part is registering the second service

lidavidm (approved):

cpp/code/CMakeLists.txt
wjones127: Ah I can move it after 👍

cpp/source/flight.rst
```diff
@@ -83,3 +83,72 @@ Finally, we'll stop our server:
...
+   :dedent: 2
+
+Creating the client
+-------------------
+
+The Flight client implementation doesn't know about any custom gRPC services,
+so to call them you'll need to create a normal gRPC client. For the Hello World
+service, we use the HelloWorldService stub, which is provided by the compiled 
+protobuf definition. Note that we can connect to either of the two ports 
+exposed.
```
lidavidm, line 151: Now it's only one port :) Though we could call out that both services are on the same port.

cpp/source/flight.rst
```diff
@@ -83,3 +83,72 @@ Finally, we'll stop our server:
...
+.. recipe:: ../code/flight.cc CustomGrpcImpl::StartServer
+   :dedent: 2
+
+Creating the client
+-------------------
+
+The Flight client implementation doesn't know about any custom gRPC services,
+so to call them you'll need to create a normal gRPC client. For the Hello World
+service, we use the HelloWorldService stub, which is provided by the compiled 
+protobuf definition. Note that we can connect to either of the two ports
```
lidavidm, line 150: Protobuf?

cpp/code/flight.cc
wjones127: Yeah I cut that out and also cut off the first part the recipe snippet that is repetitive.

lidavidm (approved):

PR #163, opened 2022-03-15
[CI] Run apt update before apt install

lidavidm:
This is [recommended](https://docs.github.com/en/actions/using-github-hosted-runners/customizing-github-hosted-runners#installing-software-on-ubuntu-runners) by the GitHub documentation because the apt package lists may be out-of-date in the runners, leading to spurious errors.


amol- (approved):
LGTM

PR #164, opened 2022-03-16
ARROW-15713: Add proper highlighting to testcode nodes

amol-:


amol-:
@davisusanibar mind checking if this works for you too? To me it seems to tackle the issue and code snippets are finally highlighted as Java instead of as Python

PR #166, opened 2022-03-17
[C++] Add TypeVisitor example

wjones127:


cpp/code/basic_arrow.cc
```diff
@@ -63,3 +66,54 @@ arrow::Status ReturnNotOk() {
...
 
 TEST(BasicArrow, ReturnNotOk) { ASSERT_OK(ReturnNotOk()); }
+
+class TypeCountVisitor : public arrow::TypeVisitor {
+ public:
+  uint64_t nested_count;
+  uint64_t non_nested_count;
+
+  template <typename T>
+  arrow::enable_if_not_nested<T, arrow::Status> Visit(const T&) {
```
wjones127, line 76: I think I must be doing this wrong, since I am getting NotImplemented for int8:

https://github.com/apache/arrow-cookbook/runs/5588880880?check_suite_focus=true#step:6:94

Any idea on why?

wjones127: Is it using the default implementation from the parent class instead?

pitrou: Because you're not overriding the actual base method, you're defining an unrelated template method which happens to have the same name :-)

pitrou: `VisitTypeInline` may allow you to use such a pattern, because it's not based on polymorphism.

cpp/source/basic.rst
```diff
@@ -48,3 +48,29 @@ boilerplate for you.  It will run the contained expression and check the resulti
...
+To implement a TypeVisitor we have to implement a Visit method for every possible
+DataType we wish to handle. Fortunately, we can often use templates and type
+traits to make this less verbose.
+
+.. literalinclude:: ../code/basic_arrow.cc
+   :language: cpp
+   :linenos:
+   :start-at: class TypeCountVisitor
+   :end-at: };  // TypeCountVisitor
+   :caption: TypeVisitor that counts nested and non-nested types
```
pitrou, line 72: This is not a terrific example, because you can simply use the `is_nested` function in `arrow/type_traits.h` without having to use a visitor.

AFAIU, the cookbook tries to showcase how to achieve real-world-ish goals, not how to use a given API.
@westonpace @amol-  What do you think?

wjones127: Agreed. I'd like to find a better example, but haven't yet thought of one that wasn't too long. But as a first step I'd like to at least figure out how to use our API.

cpp/code/basic_arrow.cc
wjones127: Ah so if I understand right, if you subclass `TypeVisitor` you can't use templates to implement the overriding methods.

pitrou: Indeed, because you must override the exact methods. Adding `override` at the end helps ensure you didn't make a mistake when overriding, btw.

cpp/source/basic.rst
nirandaperera: IMO https://github.com/apache/arrow/pull/11886 was a really good example of a `TypeVisitor` :-)

wjones127: I've created two new examples. One is a class that generates a random record batch given a schema and row count, which showcases an approach to handling nested types.

The other is a class that can sum across columns of varied number types, which showcases how to use the type traits to avoid repetitive implementations for similar types.

I think these are better use cases that my original. The first is something I've seen in the wild (and is also inspired by our internal test functions).

cpp/code/basic_arrow.cc
```diff
@@ -63,3 +69,162 @@ arrow::Status ReturnNotOk() {
...
+  arrow::Status Visit(const arrow::Array& array) {
+    return arrow::Status::NotImplemented("Can not compute sum for array of type ",
+                                         array.type()->ToString());
+  }
+
+  template <typename ArrayType, typename T = typename ArrayType::TypeClass>
+  arrow::enable_if_number<T, arrow::Status> Visit(const ArrayType& array) {
+    for (auto value : array) {
+      if (value.has_value()) {
+        partial += (double)value.value();
```
pitrou, line 181:
```suggestion
        partial += static_cast<double>(value.value());
```

cpp/source/basic.rst
```diff
@@ -48,3 +48,58 @@ boilerplate for you.  It will run the contained expression and check the resulti
...
+  
+
+.. recipe:: ../code/basic_arrow.cc GenerateRandomData
+   :dedent: 2
+
+
+Convert Arbitrary Scalars to Variants
+-------------------------------------
+
+.. TODO: Implement converter from rows to values
```
pitrou, line 89: Is this a recipe you actually want to write soon (in this PR perhaps)? Otherwise, a JIRA is better suited rather than a TODO comment directly in the source tree, IMHO.

pitrou (commented):
Thanks for the update! I agree the new examples look less artificial. Here are some more comments.

cpp/source/basic.rst
wjones127: I'll delete this. I was initially thinking of doing three examples, but I think the two I've done seem fine.

cpp/code/basic_arrow.cc
```diff
@@ -18,6 +18,12 @@
 #include <arrow/api.h>
 #include <gtest/gtest.h>
 
+#include <random>
+
+#include "arrow/array.h"
+#include "arrow/type.h"
+#include "arrow/type_traits.h"
+#include "arrow/visit_array_inline.h"
```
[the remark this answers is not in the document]
wjones127: Okay I tested this and the only one I actually needed was `#include "arrow/visit_array_inline.h"`. I've mentioned that we should add this to the API here https://issues.apache.org/jira/browse/ARROW-15975

cpp/code/basic_arrow.cc
```diff
@@ -63,3 +69,162 @@ arrow::Status ReturnNotOk() {
...
+
+  // Default implementation
+  arrow::Status Visit(const arrow::Array& array) {
+    return arrow::Status::NotImplemented("Can not compute sum for array of type ",
+                                         array.type()->ToString());
+  }
+
+  template <typename ArrayType, typename T = typename ArrayType::TypeClass>
+  arrow::enable_if_number<T, arrow::Status> Visit(const ArrayType& array) {
+    for (auto value : array) {
```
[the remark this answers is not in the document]
wjones127: Good idea. Type is `arrow::util::optional<typename T::c_type>`.

cpp/code/basic_arrow.cc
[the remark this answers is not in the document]
pitrou: Nit: can we gather related includes together and use the same quoting convention? (angle brackets apparently)

cpp/code/creating_arrow_objects.cc
```diff
@@ -58,5 +60,92 @@ arrow::Status CreatingArraysPtr() {
...
+
+  EndRecipe("GenerateRandomData");
+  EXPECT_EQ(batch->num_rows(), 5);
+
+  return arrow::Status::OK();
+}
+
 TEST(CreatingArrowObjects, CreatingArraysTest) { ASSERT_OK(CreatingArrays()); }
 TEST(CreatingArrowObjects, CreatingArraysPtrTest) { ASSERT_OK(CreatingArraysPtr()); }
+TEST(BasicArrow, GenerateRandomData) { ASSERT_OK(GenerateRandomData()); }
```
pitrou, line 151: Nit:
```suggestion
TEST(CreatingArrowObjects, GeneratingRandomData) { ASSERT_OK(GenerateRandomData()); }
```

cpp/source/basic.rst
```diff
@@ -48,3 +48,45 @@ boilerplate for you.  It will run the contained expression and check the resulti
...
+Generalize Computations Across Arrow Types
+------------------------------------------
+
+Array visitors can be useful when writing functions that can handle multiple
+array types. However, implementing a visitor for each type individually can be
+excessively verbose. Fortunately, Arrow provides type traits that allow you to
+write templated functions to handle subsets of types. The example below
+demonstrates a table sum function that can handle any integer or floating point 
+array with only a single visitor implementation by leveraging
+:cpp:type:`arrow::enable_if_number`.
```
pitrou, line 81: A pity this reference doesn't work, but that needs a separate PR to solve on the Arrow side I suppose.

pitrou (approved):
Just some minor nits. Thanks a lot for doing this!

cpp/source/basic.rst
wjones127: Yes, I'll be adding those to the API docs in https://issues.apache.org/jira/browse/ARROW-15975

cpp/code/basic_arrow.cc
```diff
@@ -63,3 +64,71 @@ arrow::Status ReturnNotOk() {
 TEST(BasicArrow, ReturnNotOkNoMacro) { ASSERT_OK(ReturnNotOkMacro()); }
 
 TEST(BasicArrow, ReturnNotOk) { ASSERT_OK(ReturnNotOk()); }
+
+/// \brief Sum numeric values across columns
+///
+/// Only supports floating point and integral types. Does not support decimals.
+class TableSummation {
+ public:
+  double partial;
```
pitrou, line 73: Is this me, or is it uninitialized? Also, this member doesn't need to be public AFAICT.
```suggestion
  double partial = 0.0;
 public:
```

pitrou (approved):
+1, just a suggestion.

cpp/code/basic_arrow.cc
wjones127: Ah I guess that is undefined behavior in C++ if it's not set. Thanks!

wjones127:
cc @lidavidm @westonpace

PR #167, opened 2022-03-18
[R] Add list of dependencies and hint for global pandoc path in CONTRIBUTING.md

stephhazlitt:
– list dependencies
– move mention of {arrow} dependency/build to list above
– add path hint if `bookdown` cannot find `pandoc` when shipped with RStudio


thisisnic (approved):
This is great; thanks @stephhazlitt!

Issue #182, opened 2022-04-12
question/request only: any plans to add Rust to this cookbook?


yuribudilov:
Greetings
I am just wondering if idiomatic Rust will likely to join this list of languages in near future?

If not, what is best language in the cookbook to use as a baseline to "convert" to Rust flavor? C++ ?

thank you

alamb:
I don't know of any -- but it would be awesome to have one. 

Perhaps we could add it to the main arrow-rs docs via https://github.com/apache/arrow-rs/blob/master/arrow/README.md or similar. 

One idea that might be helpful is make the examples more accessable (now you need to know they exist and go digging in the repo for them): https://github.com/apache/arrow-rs/tree/master/arrow/examples

westonpace:
Would we want to add a little blurb to the main page of the cookbook?  Something like:

`Examples of using the Rust implementation can be found in https://github.com/apache/arrow-rs/tree/master/arrow/examples`

This way future readers would know where to look?

alamb:
Proposed PR: https://github.com/apache/arrow-cookbook/pull/185

PR #186, opened 2022-04-24
Clarify introductory language

alamb:
# Rationale

While working on https://github.com/apache/arrow-cookbook/issues/182, I thought that the introductory text could be improved a bit -- make the language flow better as well as remove mention of a confusing and seemingly incorrect notion that the cookbooks are the same across languages. 

# Changes
Reword introductory sentences

### DIFF
--- a/build/index.html
+++ b/build/index.html
@@ -32,10 +32,10 @@
 	<body>
 		<div id="logo"><img src="arrow.png"/></div>
 		<h1>Apache Arrow Cookbook</h1>
-		<p>The cookbook is a collection of Apache Arrow recipes for
+		<p>The cookbooks are a collection of Apache Arrow recipes for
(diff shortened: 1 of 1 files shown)


amol-:
> remove mention of a confusing and seemingly incorrect notion that the cookbooks are the same across languages.

While that's not true at the moment, that's a Goal for the cookbook. There should be a minimum set of recipes that is available on all languages and that should be presented in all cookbook. The phrase was there to make that clear to anyone willing to contribute with a new cookbook. At the moment we are far from there, as some cookbooks like the C++ one are very far from what is available in Python or R, but we shouldn't discard our goals just because some cookbooks haven't been able to achieve them.

alamb:
Thank you for the clarification. 

It might help to clarify the distinction between goal and reality in the introduction to avoid confusion by other readers.

Issue #206, opened 2022-05-13
[Java] Parquet reading example fails with Arrow v8.0


toddfarmer:
Changes to the Java APIs since version 7.0 result in the cookbook example [Query Data Content for File](https://arrow.apache.org/cookbook/java/dataset.html#query-data-content-for-file) no longer compiling.  The updated API appears to be used in [this test case](https://github.com/apache/arrow/blob/dc97883dee25ba8da55c7591060c44de2ea00865/java/dataset/src/test/java/org/apache/arrow/dataset/TestDataset.java#L78-L90), and the following modified cookbook code resulted in successful compilation/execution:

```java
    String uri = "file:" + System.getProperty("user.dir") + "/thirdpartydeps/parquetfiles/data1.parquet";
    ScanOptions options = new ScanOptions(/*batchSize*/ 100);
    try (
            BufferAllocator allocator = new RootAllocator();
            DatasetFactory datasetFactory = new FileSystemDatasetFactory(allocator, NativeMemoryPool.getDefault(), FileFormat.PARQUET, uri);
            Dataset dataset = datasetFactory.finish();
            Scanner scanner = dataset.newScan(options);
            VectorSchemaRoot vsr = VectorSchemaRoot.create(scanner.schema(), allocator)
    ) {
      scanner.scan().forEach(scanTask -> {
        VectorLoader loader = new VectorLoader(vsr);
        ArrowReader reader = scanTask.execute();
        try {
          while (reader.loadNextBatch()) {
            VectorSchemaRoot root = reader.getVectorSchemaRoot();
            final VectorUnloader unloader = new VectorUnloader(root);
            loader.load(unloader.getRecordBatch());
            System.out.print(vsr.contentToTSVString());
          }
        } catch (IOException ex) {
          ex.printStackTrace();
        }
      });
    } catch (Exception e) {
      e.printStackTrace();
    }
```

Issue #201, opened 2022-05-12
[R] Cookbook deployment fails due to R tests failing


lidavidm:
https://github.com/apache/arrow-cookbook/runs/6391576879?check_suite_focus=true

```
>>> Testing R Cookbook <<<

cd ./r && Rscript ./scripts/test.R

Attaching package: ‘arrow’

The following object is masked from ‘package:testthat’:

    matches

The following object is masked from ‘package:utils’:

    timestamp


Attaching package: ‘dplyr’

The following object is masked from ‘package:testthat’:

    matches

The following objects are masked from ‘package:stats’:

    filter, lag

The following objects are masked from ‘package:base’:

    intersect, setdiff, setequal, union

✔ | F W S  OK | Context

⠏ |         0 | arrays                                                          
⠴ | 1       5 | arrays                                                          
✖ | 1       8 | arrays [0.1s]
────────────────────────────────────────────────────────────────────────────────
Failure (test-arrays.R:67:3): array_fivenum works as expected
`expect_error(fivenum(arrow_array))` did not produce any warnings.
────────────────────────────────────────────────────────────────────────────────

⠏ |         0 | creating_arrow_objects                                          
✔ |         7 | creating_arrow_objects

⠏ |         0 | flight                                                          
⠏ |         0 | python                                                          
⠋ |         1 | python                                                          
⠙ |         2 | python                                                          
✔ |         2 | python [4.4s]

⠏ |         0 | reading_and_writing_data                                        
⠦ |         7 | reading_and_writing_data                                        
⠇ |         9 | reading_and_writing_data                                        
⠴ |        16 | reading_and_writing_data                                        
⠇ |        19 | reading_and_writing_data                                        
⠸ |     2  22 | reading_and_writing_data                                        
✔ |     2  23 | reading_and_writing_data [0.8s]
────────────────────────────────────────────────────────────────────────────────
Skip (test-reading_and_writing_data.R:390:1): parquet_gzip
Reason: empty test

Skip (test-reading_and_writing_data.R:418:1): dataset_gzip
Reason: empty test
────────────────────────────────────────────────────────────────────────────────

⠏ |         0 | specify_data_types_and_schemas                                  
⠦ |         7 | specify_data_types_and_schemas                                  
✔ |         8 | specify_data_types_and_schemas [0.1s]

⠏ |         0 | tables                                                          
⠋ |         1 | tables                                                          
⠴ |         6 | tables                                                          
⠋ |   2     9 | tables                                                          
[...]

PR #207, opened 2022-05-13
[Java] Parquet reading example fails with Arrow v8.0

davisusanibar:
PR to fix problems mention at: https://github.com/apache/arrow-cookbook/issues/206

### DIFF
--- a/java/source/dataset.rst
+++ b/java/source/dataset.rst
@@ -163,24 +163,36 @@ Query Data Content For File
     import org.apache.arrow.memory.RootAllocator;
     import org.apache.arrow.vector.VectorLoader;
     import org.apache.arrow.vector.VectorSchemaRoot;
-    import java.util.stream.Stream;
+    import org.apache.arrow.vector.VectorUnloader;
+    import org.apache.arrow.vector.ipc.ArrowReader;
+    import org.apache.arrow.vector.ipc.message.ArrowRecordBatch;
+
+    import java.io.IOException;
 
     String uri = "file:" + System.getProperty("user.dir") + "/thirdpartydeps/parquetfiles/data1.parquet";
     ScanOptions options = new ScanOptions(/*batchSize*/ 100);
-    try(
+    try (
         BufferAllocator allocator = new RootAllocator();
         DatasetFactory datasetFactory = new FileSystemDatasetFactory(allocator, NativeMemoryPool.getDefault(), FileFormat.PARQUET, uri);
         Dataset dataset = datasetFactory.finish();
         Scanner scanner = dataset.newScan(options);
         VectorSchemaRoot vsr = VectorSchemaRoot.create(scanner.schema(), allocator)
-    ){
-        scanner.scan().forEach(scanTask-> {
+    ) {
+        scanner.scan().forEach(scanTask -> {
             VectorLoader loader = new VectorLoader(vsr);
-            scanTask.execute().forEachRemaining(arrowRecordBatch -> {
-                loader.load(arrowRecordBatch);
-                System.out.print(vsr.contentToTSVString());
-                arrowRecordBatch.close();
-            });
+            try(ArrowReader reader = scanTask.execute()){
+                while (reader.loadNextBatch()) {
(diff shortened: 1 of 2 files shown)


davisusanibar:
Hi @lidavidm please could you help with a review.

java/source/dataset.rst
```diff
@@ -224,11 +240,19 @@ Consider that we have these files: data1: 3 rows, data2: 3 rows and data3: 250 r
         scanner.scan().forEach(scanTask-> {
             VectorLoader loader = new VectorLoader(vsr);
             final int[] count = {1};
-            scanTask.execute().forEachRemaining(arrowRecordBatch -> {
-                loader.load(arrowRecordBatch);
-                System.out.println("Batch: " + count[0]++ + ", RowCount: " + vsr.getRowCount());
-                arrowRecordBatch.close();
-            });
+            try(ArrowReader reader = scanTask.execute()){
```
toddfarmer, line 243:
```suggestion
            try (ArrowReader reader = scanTask.execute()) {
```

toddfarmer:
Should we add any comments around these code changes to highlight that the supplied example code is version (8.0.0)-specific?

lidavidm:
Maybe we should label the top level page, or perhaps include it in the page titles, or something?

Also, I wonder if we can just snapshot the 'old' cookbook on each release and label it as the '7.0.0 cookbook' or something

java/source/demo/pom.xml
```diff
@@ -21,7 +21,7 @@
     <properties>
         <maven.compiler.source>8</maven.compiler.source>
         <maven.compiler.target>8</maven.compiler.target>
-        <arrow.version>7.0.0</arrow.version>
+        <arrow.version>8.0.0</arrow.version>
```
lidavidm, line 24: We need to get this added to the list of post-release tasks and/or make it one of the things we check as part of release.

java/source/dataset.rst
```diff
@@ -275,11 +302,19 @@ In case we need to project only certain columns we could configure ScanOptions w
...
-                System.out.print(vsr.contentToTSVString());
-                arrowRecordBatch.close();
-            });
+            try(ArrowReader reader = scanTask.execute()){
+                while (reader.loadNextBatch()) {
+                    try(VectorSchemaRoot root = reader.getVectorSchemaRoot()) {
+                        final VectorUnloader unloader = new VectorUnloader(root);
+                        try(ArrowRecordBatch arrowRecordBatch = unloader.getRecordBatch()){
+                            loader.load(arrowRecordBatch);
+                            System.out.print(vsr.contentToTSVString());
```
lidavidm, line 311: nit: why not just `root.contentToTSVString()`? Is there a need to demonstrate VectorLoader/VectorUnloader here?

lidavidm (approved):

java/source/dataset.rst
davisusanibar: Thank you @toddfarmer, changed lots of places.

toddfarmer (approved):
LGTM!  Thanks @davisusanibar !

davisusanibar:
> Maybe we should label the top level page, or perhaps include it in the page titles, or something?
> 
> Also, I wonder if we can just snapshot the 'old' cookbook on each release and label it as the '7.0.0 cookbook' or something

Ticket created: https://github.com/apache/arrow-cookbook/issues/208

java/source/dataset.rst
davisusanibar: Updated

lidavidm:
Note we can't deploy cookbook updates until we get the R cookbook fixed (#201)

davisusanibar:
What about [stackoverflow question](https://stackoverflow.com/questions/72233354/how-to-read-parquet-files-into-tables-in-java-using-apache-arrow)?

lidavidm:
Do you want to just answer it?

davisusanibar:
> Do you want to just answer it?

Yes, just mention the PR in case they need to continue with their work.

lidavidm (approved):

PR #226, opened 2022-07-01
Add example for pa.Table.from_pylist

alistaire47:
Just a little PR to add an example of how to make a Table from a list of rows with [`pyarrow.Table.from_pylist()`](https://arrow.apache.org/docs/python/generated/pyarrow.Table.html#pyarrow.Table.from_pylist). 

- I put this in the same "Create Table from Plain Types" section because it already contains an example of the complementary operation of making a Table from a dict, but we can split them out if it's preferable.
- My editor stripped a little trailing whitespace elsewhere, so apologies for the whitespace diffs, but that should reduce future whitespace diffs.
- Built everything, and it looked as expected and links worked.

### DIFF
--- a/python/source/create.rst
+++ b/python/source/create.rst
@@ -42,7 +42,7 @@ be considered nulls
 
     import numpy as np
 
-    array = pa.array([1, 2, 3, 4, 5], 
+    array = pa.array([1, 2, 3, 4, 5],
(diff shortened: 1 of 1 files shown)


AlenkaF:
Thanks for adding this example to the cookbook! +1

lidavidm (approved):

PR #227, opened 2022-07-08
174: Adding recipe for custom compute functions

drin:
This recipe shows the major portions of a custom, or new, compute
function:
- defining a compute kernel
- creating a function instance
- associating the kernel with the function
- registering the function in a registry
- calling the function

The aliases are to keep the code as readable as possible, and also to frontload the various dependencies to make it obvious to the reader.

The kernel implementation here is essentially the same as the new FastHash32 function in progress of being added. The reason for this is that there are examples of much simpler functions (AbsoluteValue), and this recipe also helps the reader understand how to call other functions and how to structure non-trivial returns.

### DIFF
--- a/cpp/code/compute_fn.cc
+++ b/cpp/code/compute_fn.cc
@@ -1,0 +1,270 @@
+// ------------------------------
+// Dependencies
+
+// standard dependencies
+#include <stdint.h>
+#include <string>
+#include <iostream>
+
+// arrow dependencies
+#include <arrow/api.h>
+#include <arrow/compute/api.h>
+#include <arrow/compute/exec/key_hash.h>
+
+#include "common.h"
+
+
+// >> aliases for types in standard library
+using std::shared_ptr;
+using std::vector;
+
+// arrow util types
+using arrow::Result;
+using arrow::Status;
+using arrow::Datum;
+
+// arrow data types and helpers
+using arrow::UInt32Builder;
+using arrow::Int32Builder;
+
+using arrow::Array;
+using arrow::ArraySpan;
+
+
+// aliases for types used in `NamedScalarFn`
+//    |> kernel parameters
+using arrow::compute::KernelContext;
+using arrow::compute::ExecSpan;
+using arrow::compute::ExecResult;
+
+//    |> other context types
+using arrow::compute::ExecContext;
+using arrow::compute::LightContext;
+
+//    |> common types for compute functions
+using arrow::compute::FunctionRegistry;
+using arrow::compute::FunctionDoc;
+using arrow::compute::InputType;
+using arrow::compute::OutputType;
+using arrow::compute::Arity;
+
+//    |> the "kind" of function we want
+using arrow::compute::ScalarFunction;
+
+//    |> structs and classes for hashing
+using arrow::util::MiniBatch;
+using arrow::util::TempVectorStack;
+
+using arrow::compute::KeyColumnArray;
+using arrow::compute::Hashing32;
+
+//    |> functions used for hashing
+using arrow::compute::ColumnArrayFromArrayData;
+
+
+// ------------------------------
+// Structs and Classes
+
+// >> Documentation for a compute function
+/**
+ * Create a const instance of `FunctionDoc` that contains 3 attributes:
+ *  1. Short description
+ *  2. Long  description (limited to 78 characters)
+ *  3. Name of input arguments
+ */
+const FunctionDoc named_scalar_fn_doc {
+   "Unary function that calculates a hash for each row of the input"
+  ,"This function uses an xxHash-like algorithm which produces 32-bit hashes."
+  ,{ "input_array" }
+};
+
+
+// >> Kernel implementations for a compute function
+/**
+ * Create implementations that will be associated with our compute function. When a
+ * compute function is invoked, the compute API framework will delegate execution to an
+ * associated kernel that matches: (1) input argument types/shapes and (2) output argument
+ * types/shapes.
+ *
+ * Kernel implementations may be functions or may be methods (functions within a class or
+ * struct).
+ */
+struct NamedScalarFn {
+
+  /**
+   * A kernel implementation that expects a single array as input, and outputs an array of
+   * uint32 values. We write this implementation knowing what function we want to
+   * associate it with ("NamedScalarFn"), but that association is made later (see
+   * `RegisterScalarFnKernels()` below).
+   */
+  static Status
+  Exec(KernelContext *ctx, const ExecSpan &input_arg, ExecResult *out) {
+    StartRecipe("DefineAComputeKernel");
+
+    if (input_arg.num_values() != 1 or not input_arg[0].is_array()) {
+      return Status::Invalid("Unsupported argument types or shape");
+    }
+
+    // >> Initialize stack-based memory allocator with an allocator and memory size
+    TempVectorStack stack_memallocator;
+    auto            input_dtype_width = input_arg[0].type()->bit_width();
+    if (input_dtype_width > 0) {
+      ARROW_RETURN_NOT_OK(
+        stack_memallocator.Init(
+           ctx->exec_context()->memory_pool()
+          ,input_dtype_width * max_batchsize
+        )
+      );
+    }
+
+    // >> Prepare input data structure for propagation to hash function
+    // NOTE: "start row index" and "row count" can potentially be options in the future
+    ArraySpan hash_input    = input_arg[0].array;
+    int64_t   hash_startrow = 0;
+    int64_t   hash_rowcount = hash_input.length;
+    ARROW_ASSIGN_OR_RAISE(
+       KeyColumnArray input_keycol
+      ,ColumnArrayFromArrayData(hash_input.ToArrayData(), hash_startrow, hash_rowcount)
+    );
+
+    // >> Call hashing function
+    vector<uint32_t> hash_results;
+    hash_results.resize(hash_input.length);
+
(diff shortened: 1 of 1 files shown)


lidavidm:
Thanks for adding this, I'll take a look.

Side note: we should probably just set up clang-format for C++ examples…

cpp/code/compute_fn.cc
```diff
@@ -0,0 +1,270 @@
...
+
+
+// ------------------------------
+// Structs and Classes
+
+// >> Documentation for a compute function
+/**
+ * Create a const instance of `FunctionDoc` that contains 3 attributes:
+ *  1. Short description
+ *  2. Long  description (limited to 78 characters)
```
lidavidm, line 72: 78 characters wide right?

cpp/code/compute_fn.cc
```diff
@@ -0,0 +1,270 @@
...
+
+  /**
+   * A kernel implementation that expects a single array as input, and outputs an array of
+   * uint32 values. We write this implementation knowing what function we want to
+   * associate it with ("NamedScalarFn"), but that association is made later (see
+   * `RegisterScalarFnKernels()` below).
+   */
+  static Status
+  Exec(KernelContext *ctx, const ExecSpan &input_arg, ExecResult *out) {
+    StartRecipe("DefineAComputeKernel");
```
lidavidm, line 102: The StartRecipe/EndRecipe might be a little too limiting in this case, since presumably we want to show the entire struct

cpp/code/compute_fn.cc
```diff
@@ -0,0 +1,270 @@
...
+    TempVectorStack stack_memallocator;
+    auto            input_dtype_width = input_arg[0].type()->bit_width();
+    if (input_dtype_width > 0) {
+      ARROW_RETURN_NOT_OK(
+        stack_memallocator.Init(
+           ctx->exec_context()->memory_pool()
+          ,input_dtype_width * max_batchsize
+        )
+      );
+    }
```
lidavidm, line 118: Wouldn't this fail for 0-width types (if that's even a thing)? Or really, is the conditional really necessary?

cpp/code/compute_fn.cc
```diff
@@ -0,0 +1,270 @@
...
+
+    // >> Prepare input data structure for propagation to hash function
+    // NOTE: "start row index" and "row count" can potentially be options in the future
```
lidavidm, line 121: You would just slice the input (0-copy). Scalar functions have to provide a row of output per row of input so such options wouldn't make sense.

cpp/code/compute_fn.cc
```diff
@@ -0,0 +1,270 @@
...
+    ArraySpan hash_input    = input_arg[0].array;
+    int64_t   hash_startrow = 0;
+    int64_t   hash_rowcount = hash_input.length;
+    ARROW_ASSIGN_OR_RAISE(
+       KeyColumnArray input_keycol
+      ,ColumnArrayFromArrayData(hash_input.ToArrayData(), hash_startrow, hash_rowcount)
+    );
```
lidavidm, line 128: While for an actual implementation, we would want to reuse Arrow hashing code, I wonder if this example would be clearer/more focused if we implemented a hash function inline (even a fairly trivial one) and removed the use of semi-internal APIs

lidavidm (commented):
Thanks, this is a great example. I left some comments, but as noted we also need an reST file to get this to appear in the Cookbook itself. Of course, this probably needs to wait on 9.0.0 too

cpp/code/compute_fn.cc
drin: yeah. I didn't realize this at the time, so I'll update it

cpp/code/compute_fn.cc
drin: You may be right. I haven't played around enough with a variety of data types, but I think I did this so that I could handle variable length types differently, which I don't even have here. I should have time to play around with it later today or tomorrow.

drin:
thanks for the comments! I will get around to the reST file today. I was putting it off a bit but today feels like a good day for it.

cpp/code/compute_fn.cc
drin: I guess this recipe can just be specific to a single data type for conciseness, so I'll remove this. I'll maybe make another recipe that shows how to work with various data types.

cpp/code/compute_fn.cc
drin: I'll look at how the Start/End Recipe functions work and figure out how to include the struct itself

drin:
rebased to try and build the cookbooks, but I am not sure how to build the documentation for the C++ code

drin:
I am not sure how to add whole function definitions to the recipe instead of portions of the body.

For example: https://github.com/apache/arrow-cookbook/blob/411d471fcecd41ee6d93c295940108c6bc76e5df/cpp/code/compute_fn.cc#L88-L90

cpp/source/compute.rst
```diff
@@ -0,0 +1,151 @@
...
+
+.. recipe:: ../code/compute_fn.cc InvokeByCallFunction
+  :caption: Use CallFunction() to invoke a compute function by name
+  :dedent: 2
+
+.. note::
+    This method allows us to specify arguments as a vector and a custom ExecContext.
+
+If an `ExecContext` is not passed to `CallFunction` (it is null), then the default
+FunctionRegistry will be used to call the function from.
```
ksuarez1423, line 46: Referring to the default FunctionRegistry here makes me wonder if this should be the other way around -- adding functions then invoking them. Or even if these should be two separate cookbook entries -- not quite sure, just want the idea to float around.

cpp/source/compute.rst
```diff
@@ -0,0 +1,151 @@
...
+
+2. Associate the kernels with a function object
+
+3. Add the function object to a function registry
+
+
+Define Function Kernels
+-----------------------
+
+A kernel is a particular function that implements desired logic for a compute function.
```
ksuarez1423, line 72: The information in this paragraph feels better suited to a tutorial -- if we just want to establish how to define an execution kernel, then I think the next paragraph is enough, but if an initialization kernel is necessary, that should also be an entry here.

cpp/source/compute.rst
```diff
@@ -0,0 +1,151 @@
...
+is wrapped in an `ArrayData` instance and referenced by `out`.
+
+
+Associate Kernels with a Function
+---------------------------------
+
+The process of adding kernels to a compute function is easy: (1) create an appropriate
+`Function` instance--`ScalarFunction` in this case--and (2) call the `AddKernel` function.
+The more difficult part of this process is repeating for the desired data types and
+knowing how the signatures work.
```
ksuarez1423, line 106: For a cookbook, I don't think we need to talk about what's easy or hard -- just put forth what to do.

cpp/source/compute.rst
```diff
@@ -0,0 +1,151 @@
...
+recipes provided in this section show some approaches to define a compute function as well
+as how to invoke a compute function by name, given a registry.
+
+
+.. contents::
+
+Invoke a Compute Function
+=========================
+
+When invoking a compute function, the function must exist in a function registry. In this
```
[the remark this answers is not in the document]
drin: oh yeah, I guess I didn't think about that being where it is accessed from. I think it's a good note

cpp/source/compute.rst
drin: yeah, my code adds it then invokes it, but I also wanted to conceptually de-couple the 2.

I think separate cookbook entries would have a lot of overlap, but I suppose it would make the concepts more accessible.

For now, I can maybe re-word to mention that an empty ExecContext actually causes a default one to be used, and where I mention the link between ExecContext and FunctionRegistry above, I can say that the default ExecContext references the default FunctionRegistry

ksuarez1423: Overlap is fine -- the point of a cookbook is, in the end, to let people who just need to grab one snippet of concept or code be able to get it without reading the whole tutorial again. 

But yes, for now, we can just mention that there's a FunctionRegistry, and in part 1, you're using the default one.

cpp/source/compute.rst
drin: mmmm.. I wrote the paragraph to help contextualize what the recipe covers and what it doesn't. Also to implicitly hint at why the kernel function is named `Exec` and clarify that both an execution kernel and intialization kernel are basically just functions associated with "the compute function". Also that they are associated as a pair, so initialization and execution kernels are not dispatched individually.

I definitely see your point, so I don't know if I should just move this to some other documentation which may not yet exist or if I should leave it here to be moved later.

ksuarez1423: I think the answer is honestly that it needs to be in some documentation which does not exist yet. Which means maybe it shouldn't be here, and we should just note down this is vital moving forward. I'm uncertain.

cpp/code/compute_fn.cc
drin: for now I just tried to use literal includes. it still references the file, and then I just set line numbers. Will experiment with the recipe stuff again in the future.

drin:
@wjones127 or @lidavidm if either of you have time, could you check this out?

I wasn't able to locally confirm that code excerpt directives worked correctly, but I am hoping to reference this cookbook from documentation for authoring compute kernels

lidavidm:
I kicked off CI, but not sure if I'll have time to look in more detail right now

drin:
I am having trouble setting up a build to figure out why it couldn't find the recipe, so I'm just going to shelve it for a bit.

Issue #228, opened 2022-07-13
[Java] JDBC adapter recipes


davisusanibar:
- JDBC adapter row format to columnar
- JDBC adapter to handling arrays

PR #229, opened 2022-07-13
[Java] Adding Arrow Java JDBC adapter examples

davisusanibar:
https://github.com/apache/arrow-cookbook/issues/228


java/source/io.rst
```diff
@@ -444,6 +444,176 @@ Reading Parquet File
...
+
+    import java.io.IOException;
+    import java.sql.ResultSet;
+    import java.sql.SQLException;
+
+    try (BufferAllocator allocator = new RootAllocator()) {
+        JdbcConnectionPool pool = JdbcConnectionPool.create(
+                "jdbc:h2:zip:./thirdpartydeps/database/jdbc-cookbook.zip!/test",
+                "sa", "");
+        ResultSet resultSet = pool.getConnection().createStatement().executeQuery(
```
lidavidm, line 477: ResultSet needs to be closed, right?

java/source/io.rst
```diff
@@ -444,6 +444,176 @@ Reading Parquet File
...
+.. testoutput::
+
+    INT_FIELD1    BOOL_FIELD2    DOUBLE_FIELD7
+    101    true    5.6478356785345E10
+    102    true    5.6478356785345E10
+    103    true    5.6478356785345E10
+
+JdbcToArrow could be created also with a custom configurations needed with the
+support of `JdbcToArrowConfig <https://arrow.apache.org/docs/java/reference/org/apache/arrow/adapter/jdbc/JdbcToArrowConfig.html>`_
+, it is useful at the moment to work with array columns.
```
lidavidm, line 501:
```suggestion
JdbcToArrow accepts configuration through `JdbcToArrowConfig
<https://arrow.apache.org/docs/java/reference/org/apache/arrow/adapter/jdbc/JdbcToArrowConfig.html>`_.
For example, the type of the elements of array columns can be specified.
```

lidavidm: We talk about arrays and row counts here, what about decimals and other things Todd added?

java/source/io.rst
```diff
@@ -444,6 +444,176 @@ Reading Parquet File
...
+                resultSet, allocator)) {
+            while (iterator.hasNext()) {
+                VectorSchemaRoot root = iterator.next();
+                System.out.print(root.contentToTSVString());
+            }
+        }
+    } catch (SQLException throwables) {
+        throwables.printStackTrace();
+    } catch (IOException e) {
+        e.printStackTrace();
```
lidavidm, line 489:
```suggestion
    } catch (SQLException | IOException e) {
        e.printStackTrace();
```
and ditto below

java/source/io.rst
```diff
@@ -444,6 +444,176 @@ Reading Parquet File
...
+
+    import java.io.IOException;
+    import java.sql.ResultSet;
+    import java.sql.SQLException;
+    import java.sql.Types;
+    import java.util.HashMap;
+
+    try (BufferAllocator allocator = new RootAllocator()) {
+        JdbcConnectionPool pool = JdbcConnectionPool.create(
+                "jdbc:h2:zip:./thirdpartydeps/database/jdbc-cookbook.zip!/test",
```
lidavidm, line 524: is an in-memory database not possible? Apache Derby lets us do that (that's what I've been using for testing in ADBC, and it's what Flight SQL uses)

java/source/io.rst
davisusanibar: > We talk about arrays and row counts here, what about decimals and other things Todd added?

I just read that, let me add that also

java/source/io.rst
davisusanibar: Let me update to use Apache Derby also to standardize that in-memory database

lidavidm: Hmm, sticking to H2 is OK then. Didn't realize Derby lacked that

java/source/io.rst
davisusanibar: Added

java/source/demo/pom.xml
```diff
@@ -21,9 +21,17 @@
     <properties>
         <maven.compiler.source>8</maven.compiler.source>
         <maven.compiler.target>8</maven.compiler.target>
-        <arrow.version>8.0.0</arrow.version>
+        <arrow.version>9.0.0.dev191</arrow.version><!-- Prepare for 9.0.0 release -->
```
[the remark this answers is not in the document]
davisusanibar: Added configuration to define Stable/Dev version at CI script level.

But we need to talk about scope because current matrix is invoked 12 times:
```
        python-version: [ '3.9' ]
        java-version: [ '11', '17', '18' ]
        compiler-version: ['11']
        arrow-version: ['8.0.0', '9.0.0.dev405']
```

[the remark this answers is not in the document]
lidavidm: IMO it's fine to only test Java 11 (especially if that's the language level we target) (or, compile with 11 and run on 17), and again: do we care about stable/nightly cookbooks? That is a separate discussion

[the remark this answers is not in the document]
davisusanibar: If we consider OS it is invoked 24 times: 12 Linux / 12 MacOS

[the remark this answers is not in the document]
davisusanibar: > IMO it's fine to only test Java 11 (especially if that's the language level we target) (or, compile with 11 and run on 17), and again: do we care about stable/nightly cookbooks? That is a separate discussion

Ok. PoC about how to pass variables matrix from .ci github to Java cookbook is possible, now we need to discuss: `do we care about stable/nightly cookbooks? `

This is the current setup:
```
    strategy:
      matrix:
        python-version: [ '3.9' ]
        java-version: [ '11', '17', '18' ]
        compiler-version: ['11']
        arrow-version: ['9.0.0.dev405']
```

This could be some capabilities that is needed to implement if we need to care about stable/dev:
- Configure skip/consider cookbook per Arrow Java version needed (for new dev features that is not supported on stable version)

lidavidm (commented):
I'm not a fan of this. We should have a full discussion about how we want to version the cookbooks.

In particular:
- The supported Arrow version needs to be made clear in the cookbook.
- This now unconditionally uses the nightly Arrow version. I don't think we should do that.
- This is now harder to build for anyone trying to run things locally, since the pom.xml is invalid by default.
- I agree we don't need to test so many Java versions.

I would rather hold this until 9.0.0 is released if we aren't going to rethink our versioning approach.

lidavidm:
This can be updated for 9.0.0 now I think

davisusanibar:
> This can be updated for 9.0.0 now I think

Updated

java/source/io.rst
```diff
@@ -461,6 +461,292 @@ Reading Parquet File
...
+There is a configuration about precision & scale for column data type needed
+(i.e. ``JdbcFieldInfo(Types.DECIMAL, /*precision*/ 20, /*scale*/ 7))``) but this
+configuration required exact matching of every row to the established scale
+for the column, and throws ``UnsupportedOperationException`` when there is a mismatch,
+aborting ResultSet processing,
+
+In this example we have BigInt data type configured on H2 Database, this is
+converted to (``/*scale*/0)`` by default, then we have a ``/*scale*/7`` configured on
+our code, this will be the error message for these differences: ``Caused by: java.lang.UnsupportedOperationException: BigDecimal scale must equal that in the Arrow vector: 0 != 7``
+if not applying ``setBigDecimalRoundingMode``
```
lidavidm, line 676:
```suggestion
In this example, we have a BigInt column. By default, the inferred scale
is 0. We override the scale to 7 and then set a RoundingMode to convert
values to the given scale.
```

java/source/io.rst
```diff
@@ -461,6 +461,292 @@ Reading Parquet File
 
 Please check :doc:`Dataset <./dataset>`
 
+Reading JDBC ResultSets
```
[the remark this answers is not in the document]
davisusanibar: Ok, lets move it to their own file

java/ext/javadoctest.py
```diff
@@ -37,6 +37,9 @@ def compile(
                 "-q",
                 "dependency:build-classpath",
                 "-DincludeTypes=jar",
+                "-Dmaven.compiler.source="+os.environ.get('compiler_version', '8'),
+                "-Dmaven.compiler.target="+os.environ.get('compiler_version', '8'),
+                "-Darrow.version="+os.environ.get('arrow_version', '9.0.0'),
```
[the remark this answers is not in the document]
davisusanibar: Yes, it was removed, this part is to pass the CI variables to the Java process and assign values if they are not set as for example local running

java/source/demo/pom.xml
```diff
@@ -27,13 +25,11 @@
       </extension>
     </extensions>
     </build>
-
     <properties>
-        <maven.compiler.source>8</maven.compiler.source>
-        <maven.compiler.target>8</maven.compiler.target>
-        <arrow.version>9.0.0</arrow.version>
+        <maven.compiler.source>injected_by_ci</maven.compiler.source>
```
lidavidm, line 29: I would still much rather these have a default version so that it can be easily used locally.

java/source/jdbc.rst
```diff
@@ -0,0 +1,309 @@
...
+.. "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+.. KIND, either express or implied.  See the License for the
+.. specific language governing permissions and limitations
+.. under the License.
+
+.. _arrow-jdbc:
+
+====
+Jdbc
+====
```
lidavidm, line 22:
```suggestion
==================
Arrow JDBC Adapter
==================
```

java/source/jdbc.rst
```diff
@@ -0,0 +1,309 @@
...
+   INT_FIELD1    BOOL_FIELD2    BIGINT_FIELD5    CHAR_FIELD16    LIST_FIELD19
+   101    true    1000000000300    some char text      [1,2,3]
+   102    true    100000000030    some char text      [1,2]
+   103    true    10000000003    some char text      [1]
+
+Configuring batch size
+======================
+
+The maximum rowCount to read each time is configured by default in 1024. This
+can be customized by setting values as needed by ``setTargetBatchSize``.
```
lidavidm, line 156:
```suggestion
By default, the adapter will read up to 1024 rows in a batch. This
can be customized via ``setTargetBatchSize``.
```

java/source/jdbc.rst
```diff
@@ -0,0 +1,309 @@
...
+vector``will be thrown during conversion.
+
+This can happen because Arrow infers the type from the ResultSet metadata, which
+is not accurate for all database drivers. The JDBC adapter lets you avoid this
+by either overriding the decimal scale, or by providing a RoundingMode via
+``setBigDecimalRoundingMode`` to convert values to the expected scale.
+
+In this example, we have a BigInt column. By default, the inferred scale
+is 0. We override the scale to 7 and then set a RoundingMode to convert
+values to the given scale.
```
lidavidm, line 237:
```suggestion
By default, the scale of any decimal values must exactly match the defined
scale of the Arrow type of the column, or else an UnsupportedOperationException
will be thrown with a message like ``BigDecimal scale must equal that in the Arrow
vector``.

This can happen because Arrow infers the type from the ResultSet metadata, which
is not accurate for all database drivers. The JDBC adapter lets you avoid this
by either overriding the decimal scale, or by providing a RoundingMode via
``setBigDecimalRoundingMode`` to convert values to the expected scale.

In this example, we have a BigInt column. By default, the inferred scale
is 0. We override the scale to 7 and then set a RoundingMode to convert
values to the given scale.
```

lidavidm (changes requested):

.github/workflows/test_java_osx_cookbook.yml
```diff
@@ -37,6 +37,8 @@ jobs:
       matrix:
         python-version: [ '3.9' ]
         java-version: [ '11', '17', '18' ]
+        compiler-version: ['11']
+        arrow-version: ['9.0.0']
```
lidavidm, line 41: I still don't understand why we're parameterizing this in CI? It just makes it harder to update Arrow.

davisusanibar: The main reason is to handle that as a matrix, this support pass an arrays of arrow_versions needed on demand, I could suggest that new changes related to a version would be handled by this CI file without change main code.

lidavidm: I don't think we should do this until we decide how we want to approach multiple Cookbook versions.

lidavidm: At the very least, I'd rather we put these CI changes separately instead of combining them with a larger change

davisusanibar: Make sense, let me replace with the last CI configuration

PR #230, opened 2022-07-15
MINOR: update copyright year to 2022

nealrichardson:
Also, it appears my editor ran `black` on these files when I saved them, so there are a few style changes too. Happy to back those out if anyone feels strongly about it.

### DIFF
--- a/cpp/source/conf.py
+++ b/cpp/source/conf.py
@@ -28,14 +28,15 @@
 #
 import os
 import sys
+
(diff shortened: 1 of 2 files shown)


lidavidm (approved):
If anything I would love to have more automated formatting (though it would only help C++ code examples as the others are all embedded within reST, maybe something like https://github.com/psf/black/issues/294 would help)

PR #232, opened 2022-07-27
[Java] Added missing semicolon in Working with Schema page

mystic-lama:
Minor fix - added missing semicolon for create Fields section. 

The code passes with make javatest, however found the issue when copy pasting into editor showed error.


mystic-lama:
@lidavidm Kindly review whenever you get 30 seconds

lidavidm (approved):
Thanks for catching

PR #233, opened 2022-07-27
[R] Recipe for nested json

eitsupi:
Related to [ARROW-17143](https://issues.apache.org/jira/browse/ARROW-17143)

json files often have a hierarchical structure, but it is surprisingly easy to convert such a json file into a flat data frame by reading it as a data frame with `read_json_arrow` function and then processing it with the `tidyr` functions.

### DIFF
--- a/r/content/reading_and_writing_data.Rmd
+++ b/r/content/reading_and_writing_data.Rmd
@@ -304,6 +304,39 @@ test_that("read_json_arrow chunk works as expected", {
 unlink(tf)
 ```
 
+### Discussion
+
+When reading a nested json file, a flat data frame can be easily created by using the `tidyr` functions.
+
+```{r, read_json_arrow1}
+library(tidyr, warn.conflicts = FALSE)
+tf <- tempfile()
+writeLines('
+    { "hello": 3.5, "world": false, "foo": { "bar": [ 1, 2 ] } }
+    { "hello": 3.25, "world": null }
+    { "hello": 0.0, "world": true, "foo": { "bar": [ 3, 4, 5 ] } }
+  ', tf, useBytes = TRUE)
+nested_data <- read_json_arrow(tf)
+unnested_data <- nested_data %>%
+  unnest(foo) %>%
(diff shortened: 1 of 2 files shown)


r/content/reading_and_writing_data.Rmd
```diff
@@ -304,6 +304,39 @@ test_that("read_json_arrow chunk works as expected", {
 unlink(tf)
 ```
 
+### Discussion
```
dragosmg, line 307: Thanks for submitting the PR. The content is great, though I'm not entirely convinced it belongs in the Cookbook.

FWIW I think this might be going into too much detail regarding working with JSON imported a list-columns in a dplyr-tidyr workflow. We could say something along the lines of ... once you import / read the data in, you can then work with it in several ways: tidyr or purrr workflows, etc. IMHO working with nested structures is slightly out-of-scope for the cookbook. We could link to the working with nested data tidyr [vignette](https://tidyr.tidyverse.org/articles/nest.html) or Jenny Bryan's working with [list-columns in purrr](https://jennybc.github.io/purrr-tutorial/ls13_list-columns.html).

thisisnic: Agreed - there's a really delicate balance to be struck here between "explaining how to do things using the arrow package" (desirable) and "explaining how to do things using the tidyverse in general" (undesirable).  We've come up against this in the past, when [writing the recipe for using dplyr verbs in arrow](https://arrow.apache.org/cookbook/r/manipulating-data---tables.html#use-dplyr-verbs-in-arrow), and making a conscious choice _not_ to detail each of the main dplyr verbs that we support (as this is dplyr content and not arrow content).

I think you're right @eitsupi that it could be helpful to point readers in the right direction here but agree with @dragosmg that this level of detail is a little out-of-scope and that a brief sentence with a link would be best in this particular case.

eitsupi: Thank you both for your review. I agree with your remarks and close this.

PR #235, opened 2022-07-29
[Java] Added section for converting Avro To Arrow

mystic-lama:
Added example to convert Avro to Arrow


mystic-lama:
@lidavidm Can use your help with this PR. Still at beginning phases of learning Arrow

lidavidm:
@davisusanibar would you like to review as well?

davisusanibar:
I have a question about how to organize our recipes for dataset such as: parquet / avro / orc

With this PR this will be the result:
image

What about if we could organize dataset cookbooks on this way?
image

Please if you could help me on this validation @lidavidm / @lwhite1

lidavidm:
The Avro adapter isn't part of datasets.

java/source/avro.rst
```diff
@@ -0,0 +1,67 @@
...
+
+    BinaryDecoder decoder = new DecoderFactory().binaryDecoder(new FileInputStream("./thirdpartydeps/avro/users.avro"), null);
+
+    Schema schema = new Schema.Parser().parse(new File("./thirdpartydeps/avro/user.avsc"));
+    AvroToArrowVectorIterator avroToArrowVectorIterator = AvroToArrow.avroToArrowIterator(schema, decoder, config);
+
+    while(avroToArrowVectorIterator.hasNext()) {
+        VectorSchemaRoot root = avroToArrowVectorIterator.next();
+        System.out.print(root.contentToTSVString());
+    }
```
davisusanibar, line 61:
```suggestion
    import org.apache.arrow.AvroToArrow;
    import org.apache.arrow.AvroToArrowConfig;
    import org.apache.arrow.AvroToArrowConfigBuilder;
    import org.apache.arrow.AvroToArrowVectorIterator;
    import org.apache.arrow.memory.BufferAllocator;
    import org.apache.arrow.memory.RootAllocator;
    import org.apache.arrow.vector.VectorSchemaRoot;
    import org.apache.avro.Schema;
    import org.apache.avro.io.BinaryDecoder;
    import org.apache.avro.io.DecoderFactory;

    import java.io.File;
    import java.io.FileInputStream;
    import java.io.FileNotFoundException;
    import java.io.IOException;

    try {
        BinaryDecoder decoder = new DecoderFactory().binaryDecoder(new FileInputStream("./thirdpartydeps/avro/users.avro"), null);
        Schema schema = new Schema.Parser().parse(new File("./thirdpartydeps/avro/user.avsc"));
        try (BufferAllocator allocator = new RootAllocator()) {
            AvroToArrowConfig config = new AvroToArrowConfigBuilder(allocator).build();
            try (AvroToArrowVectorIterator avroToArrowVectorIterator = AvroToArrow.avroToArrowIterator(schema, decoder, config)) {
                while(avroToArrowVectorIterator.hasNext()) {
                    try (VectorSchemaRoot root = avroToArrowVectorIterator.next()) {
                        System.out.print(root.contentToTSVString());
                    }
                }
            }
        }
    } catch (FileNotFoundException e) {
        e.printStackTrace();
    } catch (IOException e) {
        e.printStackTrace();
    }
```

davisusanibar:
> The Avro adapter isn't part of datasets.

It means we could continue as it is now (maybe change title from 'Avro to Arrow' to 'Avro' only?)

java/source/avro.rst
mystic-lama: @davisusanibar This issue seems to be common theme from my side. How do I ensure I don't make the same mistake again?

lidavidm:
Yeah, if we had other formats, they could be their own section, but it's just Avro and Arrow IPC itself

java/source/avro.rst
```diff
@@ -0,0 +1,75 @@
...
+
+Avro encoded data can be converted into Arrow format.
+
+.. contents::
+
+Avro to Arrow
+=============
+
+The example assumes that the avro data doesn't contain schema in the file
+to be converted into Arrow
```
lidavidm, line 32: I honestly do not know what this means? The code sample is reading a schema it appears

lidavidm: Ah, separate files.

```suggestion
The example assumes that the Avro schema is stored separately from the Avro data itself.
```

java/source/avro.rst
```diff
@@ -0,0 +1,75 @@
...
+            AvroToArrowConfig config = new AvroToArrowConfigBuilder(allocator).build();
+            try (AvroToArrowVectorIterator avroToArrowVectorIterator = AvroToArrow.avroToArrowIterator(schema, decoder, config)) {
+                while(avroToArrowVectorIterator.hasNext()) {
+                    try (VectorSchemaRoot root = avroToArrowVectorIterator.next()) {
+                        System.out.print(root.contentToTSVString());
+                    }
+                }
+            }
+        }
+    } catch (FileNotFoundException e) {
```
lidavidm, line 65: Consolidate the catches?

lwhite1:
> The Avro adapter isn't part of datasets.

ORC isn't part of datasets either (in Java, currently)

lwhite1:
With David's suggestions, this looks good to me. 

FWIW, I find the approach of providing sufficient code for the reader to create a runnable example, without actually providing a runnable (or compilable) example a little odd, but it seems to be the standard approach for the Cookbook. And it is a little shorter :)

java/source/avro.rst
mystic-lama: Arrow data file can store schema along with data. This is in reference to that. This example shall fail in that case, hence mentioned explicitly.

java/source/avro.rst
mystic-lama: Done

lidavidm:
@lwhite1 all examples are given for JShell and are runnable in that environment (`make javatest` will actually run the code and compare the results). Maybe we need to make that more explicit.

java/source/avro.rst
lidavidm: It doesn't appear to have changed, additionally the code and output are still misaligned by one space

mystic-lama: I am just following format similar to other files, hence didn't remove extra space char. Should I align? Should we do it for other files too?

lidavidm: Oops. It should generally be aligned.

mystic-lama: Let me go ahead and fix the alignment, if it looks good we shall take care in other files when we touch them. Give me 5 mins

mystic-lama: pushed the change, one issue i noted it the html rendered output has extra space between results, not sure where it is coming from. Here is how it looks in html

name    favorite_number    favorite_color
Alyssa    256    null

Ben    7    red

lidavidm (approved):

PR #242, opened 2022-08-05
[Java] Add splicing non-zero index example

mystic-lama:
Added example for splicing from a non-zero index

### DIFF
--- a/java/source/create.rst
+++ b/java/source/create.rst
@@ -213,6 +213,13 @@ In this example, we copy a portion of the input IntVector to a new IntVector.
        TransferPair tp = vector.getTransferPair(allocator);
        tp.splitAndTransfer(0, 5);
        try (IntVector sliced = (IntVector) tp.getTo()) {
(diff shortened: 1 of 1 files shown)


mystic-lama:
@lidavidm added non-zero index example

Probably last PR for this week :) If you have any more idea's for to be included in cookbook, send my way

lidavidm:
Thank you!

I don't think this is different enough to warrant a separate example though; when I asked, I was just thinking we could modify the parameters in the original example

mystic-lama:
Got it. May be we can close this PR or do you want me to update the original example with different params?

lidavidm:
I would say, just add another block like this to the original example:

```java
       tp = vector.getTransferPair(allocator);
       tp.splitAndTransfer(2, 6);
       try (IntVector sliced = (IntVector) tp.getTo()) {
           System.out.print(sliced);
       }
```

lidavidm (approved):

PR #243, opened 2022-08-08
Added missing license

mystic-lama:
Found few more files with missing license.


mystic-lama:
@lidavidm should we scan the files once for missing License and add all of them at once or just handle them as and when we discover? I just found these, looking for some solution.

lidavidm (approved):

lidavidm:
Either works; if you're already looking at this, then it might be good just to fix them all at once

PR #252, opened 2022-09-06
[C++] Add C++ build against Arrow Nightlies version

raulcd:
This PR is the first one to be able to build the different cookbooks against the latest development version of Arrow using the Nightly built packages.
Once we can build the different cookbooks with both a stable release and the Nightly version we will be able to deploy a development version of the cookbooks.
This PR also fixes our CMake build to be compatible with the new and old versions

### DIFF
--- a/cpp/code/CMakeLists.txt
+++ b/cpp/code/CMakeLists.txt
@@ -25,12 +25,13 @@ endif()
 
 # Add Arrow and other required packages
 find_package(Arrow REQUIRED)
-
-if(${ARROW_VERSION} VERSION_GREATER "9.0.0")
-    find_package(ArrowDataset REQUIRED)
-    find_package(ArrowFlight REQUIRED)
-    find_package(Parquet REQUIRED)
+if(NOT ${ARROW_VERSION} VERSION_GREATER "9.0.0")
+  get_filename_component(ARROW_CMAKE_BASE_DIR ${Arrow_CONFIG} DIRECTORY)
+  list(INSERT CMAKE_MODULE_PATH 0 ${ARROW_CMAKE_BASE_DIR})
 endif()
+find_package(ArrowDataset REQUIRED)
+find_package(ArrowFlight REQUIRED)
+find_package(Parquet REQUIRED)
 
 if (CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
     set(CMAKE_CXX_CLANG_TIDY "clang-tidy")
--- a/cpp/code/CMakeLists.txt
+++ b/cpp/code/CMakeLists.txt
@@ -48,7 +48,7 @@ function(RECIPE TARGET)
             common.cc
             main.cc
     )
-    if(${ARROW_VERSION} VERSION_GREATER "9.0.0")
+    if(TARGET Arrow::arrow_shared)
         set(ARROW_SHARED_NAME Arrow::arrow_shared)
     else()
         set(ARROW_SHARED_NAME arrow_shared)
--- a/cpp/code/CMakeLists.txt
+++ b/cpp/code/CMakeLists.txt
@@ -23,8 +23,14 @@ if (CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -stdlib=libstdc++")
 endif()
 
-# Add Arrow
-find_package(Arrow REQUIRED COMPONENTS dataset flight parquet)
+# Add Arrow and other required packages
+find_package(Arrow REQUIRED)
+
+if(${ARROW_VERSION} VERSION_GREATER "9.0.0")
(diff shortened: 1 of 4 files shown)


cpp/code/CMakeLists.txt
```diff
@@ -23,8 +23,14 @@ if (CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -stdlib=libstdc++")
 endif()
 
-# Add Arrow
-find_package(Arrow REQUIRED COMPONENTS dataset flight parquet)
+# Add Arrow and other required packages
+find_package(Arrow REQUIRED)
+
+if(${ARROW_VERSION} VERSION_GREATER "9.0.0")
```
kou, line 29: We always need the following `find_package()`s. So we need to remove this `if()`.

cpp/code/CMakeLists.txt
```diff
@@ -42,9 +48,14 @@ function(RECIPE TARGET)
             common.cc
             main.cc
     )
+    if(${ARROW_VERSION} VERSION_GREATER "9.0.0")
```
kou, line 51:
```suggestion
    if(TARGET Arrow::arrow_shared)
```

In general, feature based check is better than version check.

cpp/code/CMakeLists.txt
raulcd: but the following find_packages fail on the 9.0.0 version, when I use:
```
find_package(Arrow REQUIRED) 
find_package(ArrowDataset REQUIRED)
find_package(ArrowFlight REQUIRED)
find_package(Parquet REQUIRED)
```
I get the following issue:
```
CMake Error at CMakeLists.txt:28 (find_package):
  By not providing "FindArrowDataset.cmake" in CMAKE_MODULE_PATH this project
  has asked CMake to find a package configuration file provided by
  "ArrowDataset", but CMake did not find one.

  Could not find a package configuration file provided by "ArrowDataset" with
  any of the following names:

    ArrowDatasetConfig.cmake
    arrowdataset-config.cmake

  Add the installation prefix of "ArrowDataset" to CMAKE_PREFIX_PATH or set
  "ArrowDataset_DIR" to a directory containing one of the above files.  If
  "ArrowDataset" provides a separate development package or SDK, be sure it
  has been installed.

```

kou: Ah, we need to specify directry for `ArrowDatasetConfig.cmake` and so on only with 9.0.0:

```cmake
find_package(Arrow REQUIRED)
if(NOT ${ARROW_VERSION} VERSION_GREATER "9.0.0")
  get_filename_component(ARROW_CMAKE_BASE_DIR ${Arrow_CONFIG} DIRECTORY)
  list(INSERT CMAKE_MODULE_PATH 0 ${ARROW_CMAKE_BAE_DIR})
endif()
find_package(ArrowDataset REQUIRED)
find_package(ArrowFlight REQUIRED)
find_package(Parquet REQUIRED)
```

cpp/code/CMakeLists.txt
```diff
@@ -42,9 +49,14 @@ function(RECIPE TARGET)
...
+    else()
+        set(ARROW_SHARED_NAME arrow_shared)
+    endif()
     target_link_libraries(
             ${TARGET}
-            arrow_shared
+            ${ARROW_SHARED_NAME}
             arrow_dataset
             arrow_flight
             parquet
```
kou, line 62: Ah, sorry. I missed this.

CMake packages for Apache Arrow 10.0.0 have correct dependency information. So we don't need to all Arrow related CMake targets explicitly:

```cmake
if(TARGET Arrow::arrow_shared)
  target_link_libraries(${TARGET} ArrowDataset::arrow_dataset_shared ArrowFlight::arrow_flight_shared gtest)
else()
  target_link_libraries(parquet_shared arrow_shared)
  target_link_libraries(arrow_dataset_shared parquet_shared)
  target_link_libraries(arrow_flight_shared arrow_shared)
  target_link_libraries(${TARGET} arrow_dataset_shared arrow_flight_shared gtest)
endif()

```

raulcd: I had to use the `INTERFACE` keyword here:
```
        target_link_libraries(parquet_shared INTERFACE arrow_shared)
        target_link_libraries(arrow_dataset_shared INTERFACE parquet_shared)
        target_link_libraries(arrow_flight_shared INTERFACE arrow_shared)
```
otherwise I got a bunch of errors:
```
CMake Error at CMakeLists.txt:60 (target_link_libraries):
  IMPORTED library can only be used with the INTERFACE keyword of
  target_link_libraries
Call Stack (most recent call first):
  CMakeLists.txt:75 (recipe)
```

kou: Oh, sorry. Yes. I missed `INTERFACE`...

PR #253, opened 2022-09-06
WIP: [Java] Add Java build against Arrow Nightlies version

raulcd:


lidavidm:
Is this to have multiple cookbook builds, or is this purely for CI?

raulcd:
> Is this to have multiple cookbook builds, or is this purely for CI?

The idea is to serve both the development and stable version of the cookbooks in the future and be able to develop new cookbooks for both versions without having to wait for the release to happen in order to merge newer cookbooks

raulcd:
@lidavidm @davisusanibar I was waiting for the new Nightlies to be published using the new versioning (10.0.0-SNAPSHOT). Now that this was done, what do you think about this one?
I've added the test on the matrix against the nightlies only for Java 18 as I don't think the full matrix of jobs is necessary.
I've validated the jobs use the correct version, examples:
`MACOS 3.9, 18, ARROW_NIGHTLY=0` uses `Running with Arrow version: 9.0.0` --> https://github.com/apache/arrow-cookbook/actions/runs/3089352927/jobs/4996870325#step:5:118
`MACOS 3.9, 18, ARROW_NIGHTLY=1` uses ` Running with Arrow version: 10.0.0-SNAPSHOT` --> https://github.com/apache/arrow-cookbook/actions/runs/3089352927/jobs/4996870488#step:5:118

lidavidm (approved):
Thanks.

Is the plan to only add recipes on a branch or something then? Or is that still being worked out?

raulcd:
> Is the plan to only add recipes on a branch or something then? Or is that still being worked out?

This is still being worked out as I haven't send the proposal yet. At the moment I am adding jobs and documentation on how to build locally against both the stable and the development version.

My initial proposal is going to be:
* Main branch hosts the development version of the cookbooks, CI jobs (PRs against main) and deployment uses Arrow nightlies.
* Stable branch (either 9.0.0 or just stable) hosts the stable version of the cookbooks and CI (PRs against this branch) and deployment uses the latest Arrow Release
This will mean that if we want to merge a cookbook for the stable version we have to create a PR against the stable branch and also against main (or cherry-pick the commit).

Sorry I haven't had time to send the email, hopefully I can do it during the next couple of days but please, do let me know if you have any concerns.

lidavidm:
Ok, sounds good. I guess another option might be to mark recipes as 'stable' or 'dev' and only render the dev recipes on a separate branch. But we'll see how people feel

java/source/conf.py
```diff
@@ -36,7 +36,12 @@
 project = 'Apache Arrow Java Cookbook'
 copyright = '2022, Apache Software Foundation'
 author = 'The Apache Software Foundation'
-
+arrow_nightly=os.getenv("ARROW_NIGHTLY")
+if arrow_nightly and arrow_nightly != '0':
+    version = "10.0.0-SNAPSHOT"
```
davisusanibar, line 41: Another option to register `version` could be at `yml` side as a job parameter. In some cases for example will be needed run the job with custom version without change the code or PR needed

java/ext/javadoctest.py
```diff
@@ -54,6 +54,7 @@ def compile(
                 "dependency:build-classpath",
                 "-DincludeTypes=jar",
                 "-Dmdep.outputFile=.cp.tmp",
+                f"-Darrow.version={self.env.config.version}",
```
davisusanibar, line 57: For nightly version, this will be obtained from the last build version at nightlies.apache.org, could be needed to be a logic to obtain last version and use thata

raulcd: this gets read from the `conf.py` value. If it's nightly it will always be `10.0.0-SNAPSHOT` (or other X.0.0-SNAPSHOT version for future releases) with the changes  we have done to how we publish the version for the JAVA nightlies. The logic to select this version is on `java/source/conf.py`

java/source/conf.py
raulcd: @davisusanibar do you think this comment still applies with the new versioning `10.0.0-SNAPSHOT`, I suspect that if we want to test a specific version for a specific date instead of using the repository we probably have to manually download the artifacts for that specific day but this seems a pretty corner case scenario and I am not sure over-engineering for it is worth it.

davisusanibar: Not problem how it is now, thank you

raulcd:
@lidavidm do you think this could be merged? I have also this other PR: https://github.com/apache/arrow-cookbook/pull/264
The next steps are:
* Add another PR for R to use the nightly packages
* Add a scheduled daily workflow to run the development version for all cookbooks
* Move the current workflows for the development version to be run manually instead that on every commit to avoid causing issues while we don't have the stable vs development branches.
Once these are done we can create the new branch for the stable version and run workflows only for either stable or development depending on the branch. And finally update the deployment to deploy either the stable version or the development version.

lidavidm (approved):
Sorry for the delay. The plan sounds good to me.

PR #261, opened 2022-09-27
[C++] Improve C++ Cookbook

js8544:
ARROW-17851
Add ARROW_ASSIGN_OR_RAISE to utility macros.
Use auto when it improves readability.
Adopt C++ 14/17's idioms.
Add Visitor classes to visitor pattern.

### DIFF
--- a/cpp/code/basic_arrow.cc
+++ b/cpp/code/basic_arrow.cc
@@ -35,6 +35,11 @@ arrow::Status ReturnNotOkMacro() {
     if (!st.ok()) {
       return st;
     }
+    auto res = builder.Finish();
+    std::shared_ptr<arrow::Array> arr;
+    if (res.ok()) {
+      arr = std::move(res).ValueUnsafe();
+    }
     rout << "Appended -1 null values?" << std::endl;
     return arrow::Status::OK();
   };
@@ -47,14 +52,15 @@ arrow::Status ReturnNotOkMacro() {
 
 arrow::Status ReturnNotOk() {
   StartRecipe("ReturnNotOk");
-  std::function<arrow::Status()> test_fn = [] {
+  auto test_fn = [] {
     arrow::NullBuilder builder;
     ARROW_RETURN_NOT_OK(builder.Reserve(2));
     ARROW_RETURN_NOT_OK(builder.AppendNulls(-1));
+    ARROW_ASSIGN_OR_RAISE(auto arr, builder.Finish());
     rout << "Appended -1 null values?" << std::endl;
     return arrow::Status::OK();
   };
-  arrow::Status st = test_fn();
+  auto st = test_fn();
   rout << st << std::endl;
   EndRecipe("ReturnNotOk");
   EXPECT_FALSE(st.ok());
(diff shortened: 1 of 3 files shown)


cpp/code/basic_arrow.cc
```diff
@@ -70,48 +76,48 @@ TEST(BasicArrow, ReturnNotOk) { ASSERT_OK(ReturnNotOk()); }
...
+      return arrow::Status::NotImplemented("Can not compute sum for array of type ",
+                                           array.type()->ToString());
     }
-    return arrow::Status::OK();
   }
 };  // TableSummation
 
 arrow::Status VisitorSummationExample() {
   StartRecipe("VisitorSummationExample");
-  std::shared_ptr<arrow::Schema> schema = arrow::schema({
```
lidavidm, line 101: Note that in general, it was consciously chosen to write out return types to be explicit for readers who may be unfamiliar with Arrow. Some things like lambdas and initializing a builder on the stack make sense to simplify to me, but helpers like `schema` should be written out.

cpp/code/basic_arrow.cc
```diff
@@ -35,6 +35,11 @@ arrow::Status ReturnNotOkMacro() {
     if (!st.ok()) {
       return st;
     }
+    auto res = builder.Finish();
+    std::shared_ptr<arrow::Array> arr;
+    if (res.ok()) {
+      arr = std::move(res).ValueUnsafe();
+    }
```
lidavidm, line 42: This recipe is specifically trying to demonstrate how `Status` works and not how to construct an Array, really.

cpp/source/basic.rst
```diff
@@ -41,9 +41,9 @@ tedious:
   :caption: Checking the status of every function manually
   :dedent: 2
 
-The macro :c:macro:`ARROW_RETURN_NOT_OK` will take care of some of this
-boilerplate for you.  It will run the contained expression and check the resulting
-``Status`` or ``Result`` object.  If it failed then it will return the failure.
+The macros :c:macro:`ARROW_RETURN_NOT_OK` and :c:macro:`ARROW_ASSIGN_OR_RAISE` will take care of some of these
+boilerplates for you.  They will run the contained expression and check the resulting
+``Status`` or ``Result`` object.  If they failed then they will return the failure.
```
lidavidm, line 46: For clarity, I think Status and Result should be introduced separately

cpp/code/basic_arrow.cc
js8544: Hmm, the section that refers this code snippet is titled "Working with Status and Result" but there's no usage of `Result` at all. That's why I added these lines to demonstrate the usage of Result.

cpp/code/basic_arrow.cc
js8544: I see. I've reverted most `auto`s in f069a5be4d6544e26edd6735823e7d0442c87564

cpp/source/basic.rst
js8544: Sure. Should I split the section "Working with Status and Result" into two or should I keep it but move Result related stuff to a separate paragraph?

lidavidm: I think one section is OK, just have the text separate so it's clear what applies to what

lidavidm: (And separate code blocks)

cpp/code/basic_arrow.cc
lidavidm: As much as possible I think we should try to keep each code sample focused on one thing and one thing only, hence why I'd rather not add it here. And ideally we'd have code both for this approach as well as for the macro-based approach.

cpp/code/creating_arrow_objects.cc
```diff
@@ -28,7 +28,7 @@ arrow::Status CreatingArrays() {
   ARROW_RETURN_NOT_OK(builder.Append(1));
   ARROW_RETURN_NOT_OK(builder.Append(2));
   ARROW_RETURN_NOT_OK(builder.Append(3));
-  ARROW_ASSIGN_OR_RAISE(std::shared_ptr<arrow::Array> arr, builder.Finish())
+  ARROW_ASSIGN_OR_RAISE(auto arr, builder.Finish())
```
lidavidm, line 31: I'd probably want to revert these autos as well

lidavidm (commented):
I think the other issue is technically we aren't switching to C++17 until Arrow 10 so we may not want to update the cookbook with that until release

js8544:
> 

Oh right. I'll put this PR on hold for now.

lidavidm:
This should now be OK if rebased

js8544:
> This should now be OK if rebased

Thanks! I'll work on this sometime next week.

PR #267, opened 2022-10-19
[CI] Run cookbook recipes against arrow nightlies only manually and scheduled

raulcd:
Currently we are running both tests against the stable and development versions of Arrow on PRs this might cause some issues if working against the current stable version. The idea is to add a stable branch in the future and deploy that one but on the meantime we should only run on CI against arrow dev on demand or as a scheduled task.

### DIFF
--- a/.github/workflows/test_arrow_nightly_cookbook.yml
+++ b/.github/workflows/test_arrow_nightly_cookbook.yml
@@ -1,0 +1,141 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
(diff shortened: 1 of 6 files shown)


raulcd:
I have merged on my fork and validated I can manually run the workflow:
https://github.com/raulcd/arrow-cookbook/actions/runs/3281260557
The CPP job is failing but this seems to be a case where Arrow dev has changed and the recipe is not valid anymore as I can reproduce locally. We should fix that but can be done on a separate PR:
https://github.com/raulcd/arrow-cookbook/actions/runs/3281260557/jobs/5403045034

raulcd:
I've opened an issue with the failure: https://github.com/apache/arrow-cookbook/issues/268

amol-:
LGTM

PR #280, opened 2022-12-13
[Release] Update post release tasks for cookbooks

raulcd:
This PR adds some automation to create a tag for the cookbooks release and defines the process to update both main and stable branches on the cookbooks.
Once this process is available I will create the tag for 10.0.1 and a stable branch and will update the deployment scripts to deploy stable to the current cookbooks location and deploy main to a `/dev` location. I will also update the workflows to run CI for either development version or stable accordingly.

### DIFF
--- a/dev/release/01-prepare.sh
+++ b/dev/release/01-prepare.sh
@@ -1,0 +1,55 @@
+#!/usr/bin/env bash
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+set -ue
+
+SOURCE_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+if [ "$#" -ne 2 ]; then
+  echo "Usage: $0 <old_version> <new_version>"
+  exit 1
+fi
+
+. $SOURCE_DIR/utils-prepare.sh
+
+old_version=$1
+new_version=$2
+version_tag="apache-arrow-${new_version}"
+
+: ${PREPARE_DEFAULT:=1}
+: ${PREPARE_VERSION_PRE_TAG:=${PREPARE_DEFAULT}}
+: ${PREPARE_TAG:=${PREPARE_DEFAULT}}
+
+
+if [ ${PREPARE_TAG} -gt 0 ]; then
+  if [ $(git tag -l "${version_tag}") ]; then
+    echo "Delete existing git tag $version_tag"
+    git tag -d "${version_tag}"
+  fi
+fi
+
+if [ ${PREPARE_VERSION_PRE_TAG} -gt 0 ]; then
+  echo "Prepare ${new_version} on tag ${version_tag}"
+  update_versions "${old_version}" "${new_version}"
+  git commit -m "MINOR: [Release] Update versions for ${new_version}"
+fi
+
+if [ ${PREPARE_TAG} -gt 0 ]; then
+  git tag -a "${version_tag}" -m "[Release] Apache Arrow Release ${new_version}"
+fi
--- a/dev/release/02-update.sh
+++ b/dev/release/02-update.sh
@@ -38,7 +38,7 @@ if [ ${current_branch} != "main" ]; then
     exit 1
 fi
 
-if [ $(git branch -l "stable") ]; then
+if [ "$(git branch -l "stable")" != "" ]; then
     echo "Removing old stable branch locally and remotely"
     git branch -d stable
     git push -u apache --delete stable
@@ -48,7 +48,7 @@ echo "Creating and pushing new stable branch from main"
 git branch stable
 git push -u apache stable
 
-if [ $(git tag -l "${version_tag}") ]; then
+if [ "$(git tag -l "${version_tag}")" != "" ]; then
   echo "Delete existing git tag $version_tag"
   git tag -d "${version_tag}"
 fi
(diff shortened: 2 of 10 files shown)


dev/release/requirements.txt
```diff
@@ -0,0 +1 @@
+conda-lock
```
lidavidm, line 1: It feels slightly odd to use a mix of venv and Conda here (I guess it's easier to set up venv though)

dev/release/01-prepare.sh
```diff
@@ -0,0 +1,55 @@
...
+
+: ${PREPARE_DEFAULT:=1}
+: ${PREPARE_VERSION_PRE_TAG:=${PREPARE_DEFAULT}}
+: ${PREPARE_TAG:=${PREPARE_DEFAULT}}
+
+
+if [ ${PREPARE_TAG} -gt 0 ]; then
+  if [ $(git tag -l "${version_tag}") ]; then
+    echo "Delete existing git tag $version_tag"
+    git tag -d "${version_tag}"
```
lidavidm, line 43: Do we really need the tags? If so why do we bother deleting the old one?

lidavidm (approved):

dev/release/requirements.txt
raulcd: to be honest this was only added as a utility I could remove the `requirements.txt` and update the `README` to just specify that `conda-lock` must be installed in order to execute

dev/release/01-prepare.sh
raulcd: I thought it might be good to have a tag in case we want to deploy more versions of the cookbooks in the future and not only stable and dev, also to keep some history.
We only delete the new tag if it already exist, not the old one, this was only added as a guard in case we run the script twice but it could be removed if you think is not necessary.

lidavidm: Ah, I misread. Ok, sounds good to me.

dev/release/requirements.txt
lidavidm: Ok, thanks for explaining!

raulcd: Thanks! I've removed the `requirements.txt` file and updated the `README`

lidavidm (approved):

dev/release/utils-prepare.sh
```diff
@@ -0,0 +1,70 @@
...
+  popd
+
+  pushd "${ARROW_COOKBOOK_DIR}/java/source"
+  sed -i.bak -E -e \
+    "s/version = \"${base_version}\"/version = \"${next_version}\"/" \
+    conf.py
+  rm -f conf.py.bak
+
+  sed -i.bak -E -e \
+    "s/version = \"${next_version}-SNAPSHOT\"/version = \"${nightly_version_snapshot}\"/" \
```
kou, line 41: We can use multiple `-e` in one `sed`:

```suggestion
  sed -i.bak -E \
    -e "s/version = \"${base_version}\"/version = \"${next_version}\"/" \
    -e "s/version = \"${next_version}-SNAPSHOT\"/version = \"${nightly_version_snapshot}\"/" \
```

dev/release/README.md
```diff
@@ -0,0 +1,72 @@
...
+- Update the version for Java, Python and CPP cookbooks.
+- Update the conda lock files for the CPP cookbooks.
+- Create a tag for the stable Release.
+- Commit to the current branch with the updated versions.
+
+You can push the Release tag:
+
+```
+git push -u apache apache-arrow-<version>
+```
```
kou, line 62: How about pushing this tag after this commit is merged?
If this change is rejected, this tag refers nonexistent commit in `main`.

dev/release/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+And you can create a PR to merge the main branch changes.
```
kou, line 64: Note that we can't use "Squash and merge" button for this. Because it changes commit ID that is referred by the above tag.

We need to use `git push -u apache main` for the commit.

dev/release/01-prepare.sh
```diff
@@ -0,0 +1,55 @@
...
+fi
+
+if [ ${PREPARE_VERSION_PRE_TAG} -gt 0 ]; then
+  echo "Prepare ${new_version} on tag ${version_tag}"
+  update_versions "${old_version}" "${new_version}"
+  git commit -m "MINOR: [Release] Update versions for ${new_version}"
+fi
+
+if [ ${PREPARE_TAG} -gt 0 ]; then
+  git tag -a "${version_tag}" -m "[Release] Apache Arrow Release ${new_version}"
```
kou, line 54: How about running this after the update version commit is merged into `main`?

raulcd: Thanks @kou I did not think about the PR updating the commit hash. I think I will split it in two steps then. The first one will update versions and will prompt to create a PR. Once the PR is merged a second script will create the version tag from main and update stable. I prefer maintaining a PR instead of pushing to main, this will allows us to trigger CI and validate the new version did not break anything.

raulcd:
I've split the process into:
- Run bump-version script
- Create PR and merge
- Run update script to update stable and generate new tag
@kou let me know what you think about it.

dev/release/02-update.sh
```diff
@@ -0,0 +1,58 @@
...
+    echo "You should create a tag and update stable from the main branch instead of ${current_branch}:"
+    echo "1. Checkout the default branch and pull the latest changes."
+    echo "Commands:"
+    echo "   git checkout main"
+    echo "   git pull"
+    echo "   dev/release/02-update.sh ${new_version}"
+    exit 1
+fi
+
+if [ $(git branch -l "stable") ]; then
```
kou, line 41:
```suggestion
if [ "$(git branch -l "stable")" != "" ]; then
```

dev/release/02-update.sh
```diff
@@ -0,0 +1,58 @@
...
+    echo "Removing old stable branch locally and remotely"
+    git branch -d stable
+    git push -u apache --delete stable
+fi
+
+echo "Creating and pushing new stable branch from main"
+git branch stable
+git push -u apache stable
+
+if [ $(git tag -l "${version_tag}") ]; then
```
kou, line 51:
```suggestion
if [ "$(git tag -l "${version_tag}")" != "" ]; then
```

dev/release/README.md
```diff
@@ -0,0 +1,71 @@
...
+main branch. This script requires a single argument with the `new_version`:
+
+```
+./dev/release/02-update.sh 11.0.0
+```
+
+The script will:
+
+- Regenerate the stable branch from main. Take into account that this will
+delete and create a new stable branch.
```
kou, line 69:
```suggestion
- Regenerate the stable branch from main. Take into account that this will
  delete and create a new stable branch.
```

kou (approved):
+1

PR #282, opened 2022-12-14
[Dev] Configure merge to only allow Squash and merge

raulcd:
I did not find any documentation on how we merge on the Cookbooks. I thought it might be worth to add a note for committers.
I've also configured the repo to only allow `Squash and merge` to be consistent.
As reference this is the same configuration as arrow/adbc: https://github.com/apache/arrow-adbc/blob/main/.asf.yaml#L19-L22
I've also added the notification on opened and closed issues to the issues ML as we have for other repos.

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -16,6 +16,10 @@
 # under the License.
 
 github:
+  enabled_merge_buttons:
+    merge: false
+    rebase: false
+    squash: true
   features:
     issues: true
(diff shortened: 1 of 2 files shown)


raulcd:
The configuration will apply once is merged.

raulcd:
@kou what do you think about this?

kou (approved):
+1

PR #283, opened 2022-12-16
Add Matomo Analytics to cookbook

jorisvandenbossche:
Similar to what we did for the main arrow docs in https://github.com/apache/arrow/pull/13096. 

I tested it locally for the python version (and setup is exactly the same for cpp and java, all identical sphinx sites). ~I am not sure what the equivalent would be for R (since we are using bookdown and not pkgdown as for the main arrow docs, and I don't directly see the same option in bookdown)~ I have added it to R bookdown as well (based on how we did it for the main docs), but haven't tested this myself.

### DIFF
--- a/build/index.html
+++ b/build/index.html
@@ -28,6 +28,24 @@
 				width: 100%;
 			}
 		</style>
+		<!-- Matomo -->
+		<script>
+			var _paq = window._paq = window._paq || [];
+			/* tracker methods like "setCustomDimension" should be called before "trackPageView" */
+			/* We explicitly disable cookie tracking to avoid privacy issues */
+			_paq.push(['disableCookies']);
+			_paq.push(['trackPageView']);
+			_paq.push(['enableLinkTracking']);
+			(function() {
+				var u="https://analytics.apache.org/";
+				_paq.push(['setTrackerUrl', u+'matomo.php']);
+				_paq.push(['setSiteId', '20']);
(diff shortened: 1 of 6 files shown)


raulcd (commented):
I've tested on my fork both on R and Python. On Python I can see it on my deployed GitHub pages source code: https://raul.dev/arrow-cookbook/py/index.html
image
But I am not sure the bookdown part works as I can't seem to be able to find it on the generated pages:
https://raul.dev/arrow-cookbook/r/index.html
The commit I tested is this one: https://github.com/raulcd/arrow-cookbook/commit/f7b92164b85ce35f2eab2408ae730504eb50dd0a --> I copy&pasted your changes instead of cherry-picking.
On a different note, we have an `index.html` pre-constructed with the different links to all the cookbooks, should we add it there too?
https://github.com/apache/arrow-cookbook/blob/main/build/index.html

jorisvandenbossche:
@raulcd thanks for testing!

I updated index.html to include it as well (and for the specific languages moved the actual snippet to a separate file that can be included in the template, reducing the many copies of the snippet).

For R, this now matches what is included in the bookdown docs: https://bookdown.org/yihui/bookdown/yaml-options.html#yaml-options

(but if it doesn't directly work for R, I would say it is still worth to already merge for the others)

raulcd (approved):
I haven't tested the latest change but looks good to me. I agree, we can merge this and we can validate the R bookdown once is deployed as there's already value on merging as is. If the bookdown header doesn't work we can follow that up on a separate issue

PR #291, opened 2023-01-10
[Java] Make Flight example executable out of the box

liry:


lidavidm:
Hmm, these examples intentionally use JShell to cut down on some boilerplate

PR #293, opened 2023-01-27
MINOR: [Java] Guide clients to use tickets from FlightInfo

liry:
Not to construct them themselves in their clients.


lidavidm (approved):
Thanks!

PR #298, opened 2023-03-09
wip: add opentelemetry flight instructions

wjones127:


wjones127:
Huh so I realized that Arrow flight isn't built with otel support in Conda. So we can't really showcase the middleware here. I can vendor a simplified version and mention there is also one provided in the library? Or can wait on this until we get https://github.com/conda-forge/cpp-opentelemetry-sdk-feedstock/issues/38 merged.

PR #299, opened 2023-03-13
[Python][Flight] Show how to propagate opentelemetry spans to Flight

### LINKED ISSUE #297: [Python][Flight] Show how to propagate opentelemetry spans to Flight

wjones127:
Closes #297

### DIFF
--- a/python/source/flight.rst
+++ b/python/source/flight.rst
@@ -757,11 +757,17 @@ Running the client script, you should see the server printing out information ab
 Propagating OpenTelemetry Traces
 ================================
 
-OpenTelemetry_ traces can be propagated between Flight clients and servers using
-a middleware. The client middleware needs to inject the trace context into
-the call headers. The server middleware needs to extract the trace context
-from the headers and pass the context into a new span. Optionally, the client
-middleware can also create a new span to time the client-side call.
+Distributed tracing with OpenTelemetry_ allows collecting call-level performance
+measurements across a Flight service. In order to correlate spans across a Flight
+client and server, trace context must be passed between the two. This can be passed
+manually through headers in :class:`pyarrow.flight.FlightCallOptions`, or can 
+be automatically propagated using middleware.
+
+This example shows how to accomplish trace propagation through middleware.
+The client middleware needs to inject the trace context into the call headers.
+The server middleware needs to extract the trace context from the headers and
+pass the context into a new span. Optionally, the client middleware can also
+create a new span to time the client-side call.
 
 .. _OpenTelemetry: https://opentelemetry.io/docs/instrumentation/python/getting-started/
 
--- a/python/source/flight.rst
+++ b/python/source/flight.rst
@@ -771,7 +771,7 @@ create a new span to time the client-side call.
 
 .. _OpenTelemetry: https://opentelemetry.io/docs/instrumentation/python/getting-started/
 
-First, define the client middleware:
+**Step 1: define the client middleware:**
 
 .. testcode::
 
@@ -807,7 +807,7 @@ First, define the client middleware:
                 self._span.set_status(StatusCode.OK)
             self._span.end()
 
-Next, define the server middleware:
+**Step 2: define the server middleware:**
 
 .. testcode::
 
@@ -838,16 +838,20 @@ Next, define the server middleware:
                 self._span.set_status(StatusCode.OK)
             self._span.end()
 
+**Step 3: configure the trace exporter, processor, and provider:**
+
 Both the server and client will need to be configured with the OpenTelemetry SDK
 to record spans and export them somewhere. For the sake of the example, we'll 
 collect the spans into a Python list, but this is normally where you would set
-them up to be exported to some service like `Jaeger`_.
+them up to be exported to some service like `Jaeger`_. See other examples of 
+exporters at `OpenTelemetry Exporters`_.
 
 As part of this, you will need to define the resource where spans are running.
 At a minimum this is the service name, but it could include other information like
 a hostname, process id, service version, and operating system.
 
 .. _Jaeger: https://www.jaegertracing.io/
+.. _`OpenTelemetry Exporters`: https://opentelemetry.io/docs/instrumentation/python/exporters/
 
 .. testcode::
 
@@ -878,6 +882,8 @@ a hostname, process id, service version, and operating system.
         provider.add_span_processor(processor)
         trace.set_tracer_provider(provider)
         return exporter
+
+**Step 4: add the middleware to the server:**
 
 We can use the middleware now in our EchoServer from earlier. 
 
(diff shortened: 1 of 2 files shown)


wjones127:
@lidavidm any guesses on why the flight call might be cancelled? This doesn't happen for me locally.

https://github.com/apache/arrow-cookbook/actions/runs/4420556401/jobs/7750400076#step:4:162

https://github.com/apache/arrow-cookbook/blob/7d2e2e134d78c6a7844f3efe47b1072525804f4e/python/source/flight.rst?plain=1#L930-L932

lidavidm:
Because you aren't draining the result of DoAction

wjones127:
@amoeba would you be willing to review?

amoeba:
Sure, I'll take a look now

python/source/flight.rst
```diff
@@ -752,4 +752,204 @@ Running the client script, you should see the server printing out information ab
...
+            else:
+                self._span.set_status(StatusCode.OK)
+            self._span.end()
+
+Both the server and client will need to be configured with the OpenTelemetry SDK
+to record spans and export them somewhere. For the sake of the example, we'll 
+collect the spans into a Python list, but this is normally where you would set
+them up to be exported to some service like `Jaeger`_.
+
+As part of this, you also define the resource: where the spans are running.
```
amoeba, line 840:
```suggestion
As part of this, you will need to define the resource where spans are running.
```

python/source/flight.rst
```diff
@@ -752,4 +752,204 @@ Running the client script, you should see the server printing out information ab
...
+    server = EchoServer(
+        location="grpc://0.0.0.0:8816",
+        middleware={
+            "tracing": ServerTracingMiddlewareFactory()
+        },
+    )
+    t = threading.Thread(target=server.serve)
+    t.start()
+
+Finally, we can use the middle on the client. When we make a call with our
```
amoeba, line 907:
```suggestion
Finally, we can use the middleware on the client. When we make a call with our 
```

python/source/flight.rst
```diff
@@ -752,4 +752,204 @@ Running the client script, you should see the server printing out information ab
...
+        location="grpc://0.0.0.0:8816",
+        middleware={
+            "tracing": ServerTracingMiddlewareFactory()
+        },
+    )
+    t = threading.Thread(target=server.serve)
+    t.start()
+
+Finally, we can use the middle on the client. When we make a call with our 
+Flight client within an OpenTelemetry span, our client middleware will create
```
amoeba, line 908:
```suggestion
client within an OpenTelemetry span, our client middleware will create
```

Keeping terminology consistent here ("client" and "server" have been used without "Flight" prepended)

python/source/flight.rst
```diff
@@ -752,4 +752,204 @@ Running the client script, you should see the server printing out information ab
...
+a child span for the client-side flight call and then propagate the span
```
amoeba, line 909:
```suggestion
a child span for the client-side Flight call and then propagate the span
```

python/source/flight.rst
```diff
@@ -752,4 +752,204 @@ Running the client script, you should see the server printing out information ab
...
+context to the server. Our server middleware will pick up that context and create
```
amoeba, line 910: "span context" or "trace context"? I want to say the latter.

python/source/flight.rst
```diff
@@ -752,4 +752,204 @@ Running the client script, you should see the server printing out information ab
...
 .. _Arrow testing data repository: https://github.com/apache/arrow-testing/tree/master/data/flight
-.. _openssl: https://www.ibm.com/docs/en/api-connect/2018.x?topic=overview-generating-self-signed-certificate-using-openssl
\ No newline at end of file
+.. _openssl: https://www.ibm.com/docs/en/api-connect/2018.x?topic=overview-generating-self-signed-certificate-using-openssl
+
+Propagating OpenTelemetry Traces
+================================
+
+OpenTelemetry_ traces can be propagated between Flight clients and servers using
+a middleware. The client middleware needs to inject the trace information into
```
amoeba, line 761:
```suggestion
a middleware. The client middleware needs to inject the trace context into
```

I think we inject and extract "trace context" or "context" not spans. Do I have that right?

python/source/flight.rst
```diff
@@ -752,4 +752,204 @@ Running the client script, you should see the server printing out information ab
...
+the call headers. The server middleware needs to extract the trace information
```
amoeba, line 762:
```suggestion
the call headers. The server middleware needs to extract the trace context
```

python/source/flight.rst
```diff
@@ -752,4 +752,204 @@ Running the client script, you should see the server printing out information ab
...
 .. _Windows: https://docs.microsoft.com/en-us/dotnet/core/additional-tools/self-signed-certificates-guide
 .. _Arrow testing data repository: https://github.com/apache/arrow-testing/tree/master/data/flight
-.. _openssl: https://www.ibm.com/docs/en/api-connect/2018.x?topic=overview-generating-self-signed-certificate-using-openssl
\ No newline at end of file
+.. _openssl: https://www.ibm.com/docs/en/api-connect/2018.x?topic=overview-generating-self-signed-certificate-using-openssl
+
+Propagating OpenTelemetry Traces
+================================
+
+OpenTelemetry_ traces can be propagated between Flight clients and servers using
```
amoeba, line 760: Minor thing but I might start this paragraph with the problem being solved with the example. Something like:

> In order to correlate spans across a Flight client server, trace context must be passed between the two. There are two methods to do this: (1) middleware (2) manually via .... The example shown below uses a Flight middleware (hyperlink) which has the advantage of separating the implementation of the client or server from injecting and extracting trace context.

amoeba (changes requested):
This looks good @wjones127 and should be generally useful. I think I found some minor issues and also made a note or two. I didn't try running the code but it looks reasonable at a glance.

python/source/flight.rst
wjones127:
```suggestion
context to the server. Our server middleware will pick up that trace context and create
```

amoeba (approved):
Looks great, @wjones127. I like the new intro paragraphs.

python/source/flight.rst
```diff
@@ -752,4 +752,210 @@ Running the client script, you should see the server printing out information ab
...
-.. _openssl: https://www.ibm.com/docs/en/api-connect/2018.x?topic=overview-generating-self-signed-certificate-using-openssl
\ No newline at end of file
+.. _openssl: https://www.ibm.com/docs/en/api-connect/2018.x?topic=overview-generating-self-signed-certificate-using-openssl
+
+Propagating OpenTelemetry Traces
+================================
+
+Distributed tracing with OpenTelemetry_ allows collecting call-level performance
+measurements across a Flight service. In order to correlate spans across a Flight
+client and server, trace context must be passed between the two. This can be passed
```
lidavidm, line 762: Are we OK with using jargon without (briefly) defining it?

python/source/flight.rst
```diff
@@ -752,4 +752,210 @@ Running the client script, you should see the server printing out information ab
...
+
+   with tracer.start_as_current_span("hello_world") as span:
+       action = pa.flight.Action("echo", b"Hello, world!")
+       # Call list() on do_action to drain all results.
+       list(client.do_action(action=action))
+    
+   print(f"There are {len(exporter.spans)} spans.")
+   print(f"The span names are:\n  {list(span.name for span in exporter.spans)}.")
+   print(f"The span status codes are:\n  "
+         f"{list(span.status.status_code for span in exporter.spans)}.")
```
lidavidm, line 944: It may not be clear here, but the status codes here are OpenTelemetry statuses, and not gRPC or Flight statuses

python/source/flight.rst
wjones127: I think so. My interpretation of the Cookbook is that it's mainly for sharing ready to use examples, not walking the reader through all the concepts. I think by the time the user is googling "arrow flight opentelemetry python" they know what these terms mean.

lidavidm: OK, sounds good to me.

wjones127:
Added numbered steps since it seems to work well with Google search featured snippets:
Screen Shot 2023-03-16 at 11 49 03 AM

python/source/flight.rst
wjones127: We set them in the middleware above, so I think that's self explanatory enough for now. I suppose a more full-featured middleware would also set the gRPC status in the middleware with semantic conventions, but that seemed like it would overcomplicate this.

lidavidm (approved):

Issue #323, opened 2023-08-28
[C++] C++ cookbooks are failing to find zlib


raulcd:
C++ cookbooks fail with (https://github.com/apache/arrow-cookbook/actions/runs/5977593937/job/16217896402):

```
-- Could NOT find ZLIB (missing: ZLIB_LIBRARY) (found version "1.2.11")
-- Could NOT find ZLIB (missing: ZLIB_LIBRARY) (found version "1.2.11")
...
...
... OTHER NOT RELEVANT LOGS
...
...
CMake Error at /usr/share/miniconda/envs/cookbook-cpp-dev/lib/cmake/grpc/gRPCTargets.cmake:61 (set_target_properties):
  The link interface of target "gRPC::address_sorting" contains:
-- Generating done (0.0s)

    ZLIB::ZLIB

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.

Call Stack (most recent call first):
  /usr/share/miniconda/envs/cookbook-cpp-dev/lib/cmake/grpc/gRPCConfig.cmake:23 (include)
  CMakeLists.txt:81 (find_package)


CMake Error at /usr/share/miniconda/envs/cookbook-cpp-dev/lib/cmake/grpc/gRPCTargets.cmake:70 (set_target_properties):
  The link interface of target "gRPC::gpr" contains:

    ZLIB::ZLIB

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.

Call Stack (most recent call first):
  /usr/share/miniconda/envs/cookbook-cpp-dev/lib/cmake/grpc/gRPCConfig.cmake:23 (include)
  CMakeLists.txt:81 (find_package)


CMake Error at /usr/share/miniconda/envs/cookbook-cpp-dev/lib/cmake/grpc/gRPCTargets.cmake:97 (set_target_properties):
  The link interface of target "gRPC::grpc++" contains:

    ZLIB::ZLIB

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.

Call Stack (most recent call first):
  /usr/share/miniconda/envs/cookbook-cpp-dev/lib/cmake/grpc/gRPCConfig.cmake:23 (include)
[...]

PR #328, opened 2023-09-21
Fix C++ cookbook by adding zlib to conda environment files

amoeba:
Closes https://github.com/apache/arrow-cookbook/issues/323

I added zlib both to the stable and dev environment files because I assumed we'd eventually see whatever change libgrpc made in stable. I also ran conda lock --file environment.yml --kind explicit -p linux-aarch64 -p linux-64 -p osx-arm64 to update lockfiles.


raulcd (approved):
LGTM

Issue #347, opened 2024-04-11
[Java] Java Cookbook fails on 16.0.0-SNAPSHOT


amoeba:
When I run `make javatest` from the latest main with `ARROW_NIGHTLY=1`, the Flight cookbook fails with,

```
Running with Arrow version: 16.0.0-SNAPSHOT
...snip...
Exception java.lang.NoClassDefFoundError: io/grpc/BindableService
      at FlightServer.builder (FlightServer.java:169)
      at (#39:4)
Caused by: java.lang.ClassNotFoundException: io.grpc.BindableService
      at BuiltinClassLoader.loadClass (BuiltinClassLoader.java:641)
      at ClassLoaders$AppClassLoader.loadClass (ClassLoaders.java:188)
      at ClassLoader.loadClass (ClassLoader.java:525)
      ...
```

I'm not sure what might be causing this but I wonder if it's related to the recent JPMS changes. @jduo do you have any ideas? In case you're not familiar with how the Java cookbooks run, they essentially,

1. Generate a classpath using a stub pom.xml in https://github.com/apache/arrow-cookbook/blob/main/java/source/demo/pom.xml and by calling maven,
    https://github.com/apache/arrow-cookbook/blob/83aff2d04e3ef4f70895475e52e01d1c974e793c/java/ext/javadoctest.py#L53-L58
2. Runs each code snippet with jshell using the above classpath via https://github.com/apache/arrow-cookbook/blob/83aff2d04e3ef4f70895475e52e01d1c974e793c/java/ext/javadoctest.py#L80

TODOs from the thread below:

- [ ] If we keep using jshell, update the invocation to match what we have in the main arrow docs, see note in https://github.com/apache/arrow-cookbook/issues/347#issuecomment-2049622165

jduo:
I haven't seen this error.

It could be related to JPMS. We also updated gRPC's few times, merged some flight modules into one module, and changed some of the shading in Flight (though that may have only affected clients and not the server).

I'll take a peek at the cookbook and let you know how it goes @amoeba

amoeba:
Thanks so much @jduo.

jduo:
@amoeba , I was able to make this work by making some changes to the POM. This seems like the wrong way to resolve the issue though -- we seem to have some problems getting transitive dependencies.

- I changed the importing of flight-core to use the "shaded" classifier instead of the default.
- I also needed to add the following dependencies explicitly:
```
        <dependency>
              <groupId>io.netty</groupId>
[...]

PR #350, opened 2024-05-21
Refactor Java cookbooks to switch from JShell to Maven

amoeba:
In https://github.com/apache/arrow-cookbook/issues/347 we found the way we have been running cookbooks for Java (JShell) doesn't work well with JPMS which was introduced in Arrow 16. This refactors `javadoctest.py` to run examples directly with Maven using `exec:java` instead of with JShell. This PR also bumps the Java source/target version to 11 to fix some compiler errors and fixes a few compilation errors in cookbook code.

I ran into one snag will require a follow-up commit to this PR: The way the examples in [substrait.rst](https://raw.githubusercontent.com/apache/arrow-cookbook/main/java/source/substrait.rst) are written doesn't work with my approach. My approach splits each code snippet into its `import` statements and non-`import` statements, puts the imports outside the main class definition and puts the non-imports inside the class's main method. This works fine for every example except [substrait.rst](https://raw.githubusercontent.com/apache/arrow-cookbook/main/java/source/substrait.rst) which needs some of its code to be defined in the main class, e.g.,

We probably generally want to support examples that need this so I think we may need to rewrite all the Java cookbooks to have an explicit main class. @lidavidm suspected this might be the case in https://github.com/apache/arrow-cookbook/issues/347#issuecomment-2049725497 but I do wonder if there is still a way to avoid this. Any ideas welcome.


lidavidm:
Honestly, why try to support writing Java code inline in reST in the first place? It's a bad experience for recipe authors. If the examples were all external Java files that we could compile there would perhaps be some more boilerplate but it would be nothing unusual for Java and it would avoid the cookbooks being such special snowflakes.

amoeba:
The current CI failures are the expected ones (one is to bump Avro and the other is the explicit main class thing mentioned above).

amoeba:
That would probably be better in many ways, though I don't think it solves the problem of how to mix the prose and code together. Might be worth some research into what's already been built that does something like what we're aiming to do here (literate programming) for Java.

lidavidm:
https://github.com/apache/arrow-adbc/blob/main/docs/source/ext/adbc_cookbook.py is fully general, though I haven't actually tried it with Java (but it works with C++ and Python). 

There's also https://github.com/apache/arrow-adbc/blob/main/docs/source/ext/javadoc_inventory.py which allows crosslinking from Sphinx into Javadoc without hardcoding URLs.

amoeba:
Thanks @lidavidm, those look like solid options. For the purposes of this PR and getting the cookbooks working for 16.X, I think I'll just wrap the current examples in class defs and plan to refactor things in the near future.

amoeba:
I updated the Substrait cookbook code and how javadoctest.py works so all the cookbooks pass now locally when I bump Arrow to 16.0.0. I left the version bump out of this PR since we bump versions separately so CI will still fail on the Avro cookbook.

I'll file an issue to rework the cookbooks in a sec here. @raulcd do you want to take a look before I merge this?

raulcd (approved):
Maybe we could add some kind of sphinx plugin to insert Java examples from code files into the rst instead of testing the prose. Or even some templating with Jinja to generate the final rst. This probably would make much easier testing the examples.
As for the current code changes, I am not entirely sure about the test failures but if we are happy to merge now and fix those once we upgrade to the new Arrow version I am ok with that.

Issue #348, opened 2024-04-12
[Release] Update cookbook versions for 16.0.0 release


amoeba:
We can normally get away with just bumping versions in a minor PR but two issues have already come up in early testing for the 16.0.0 release so I thought I'd track those and anything else we find here ahead of the release.

- [x] Refactor Java cookbooks from jshell to maven to support post-JPMS Arrow
  - [x] https://github.com/apache/arrow-cookbook/issues/347
- [x] Update R cookbook since we now support casting to/from float16()
  - https://github.com/apache/arrow/pull/40067 added support for casting, we need to update that cookbook
  - https://github.com/apache/arrow-cookbook/actions/runs/8656358859/job/23736767598 shows the CI failure and I was able to confirm it locally

raulcd:
Let me know if I can help with anything here @amoeba

amoeba:
Hey @raulcd, this slipped. I'll work on a PR for the Java refactor this week.

amoeba:
I put a PR up at https://github.com/apache/arrow-cookbook/pull/350. I ran into one snag which may require rewriting all Java examples to have an explicit main class. I'd appreciate a review if you have a minute.

PR #357, opened 2024-10-28
GH-348: [Release] Update versions for 16.0.0

amoeba:
Closes https://github.com/apache/arrow-cookbook/issues/348

This PR:

- Updates versions from 15.0.0 to 16.0.0
- Updates the R cookbook to pass under 16.0.0
- Updates the C++ cookbook to pass under 16.0.0
- Changes `test_cpp_cookbook.yml` to now always test against stable libarrow.  This is a workaround for not having an up-to-date conda-nightlies channel, see https://github.com/apache/arrow/issues/41856


amoeba:
@raulcd whenever you have a second can you double-check my thinking here on merging this as-is? This can definitely wait until after other release tasks.

The dev version of the C++ cookbooks have been failing because the [conda-nightlies channel is broken](https://github.com/apache/arrow/issues/41856). We've talked about what to do here but I dropped the ball on figuring out a solution. In this PR I modified the C++ workflow to just test from the stable channel just so everything passes. Before this PR, it was testing against the nightly channel.

I think this a decent direction to go with for now until the nightlies channel is fixed. Testing the cookbook against the nightlies is still useful IMO so we could restore the change this PR makes when that's resolved.

raulcd:
I suppose the problem is that main should run against current dev and the stable branch against the latest release instead of a really old release like 16.0.0. I am fine having both run against the latest stable release temporarily while we find how to fix so main runs against dev. At this point I am thinking if building ourselves and or using wheels would be easier than conda nightlies.

I am ok merging this if this is an interim to publish 18.0.0. It is better than what we have already :)
Maybe we should add a note to the README to update or add a disclaimer to this:
```
See https://arrow.apache.org/cookbook/ for the latest published version using the latest stable version of Apache Arrow. See https://arrow.apache.org/cookbook/dev for the latest published version using the development version of Apache Arrow.
```

amoeba:
I added a note about the state of the dev cookbooks in https://github.com/apache/arrow-cookbook/pull/357/commits/4db967fac1738329c31fef80e3c7fffa636223d9 and will merge this now.

amoeba:
After merging, locally I ran `dev/release/02-update.sh 16.0.0` which updated the stable branch and tagged `apache-arrow-16.0.0`.

PR #371, opened 2025-01-20
MINOR: [Release] Update versions for 19.0.0

amoeba:
I ran `./dev/release/01-bump-versions.sh 18.1.0 19.0.0`

Also fixed:

- Update CMakeLists.txt to use standard CMake target name (see https://github.com/google/googletest/blob/main/googletest/README.md). I think this change is better.
- Updated Flight cookbook for Parquet changes in Arrow 19.0.0, https://github.com/apache/arrow/issues/44784


amoeba:
It looks like the Maven packages didn't upload successfully, see that https://mvnrepository.com/artifact/org.apache.arrow/arrow-memory is missing Arrow 19.0.0.

lidavidm:
I don't think we included arrow-java in the 19.0.0 release.

amoeba:
Yeah, that totally slipped my mind a few hours ago. I reverted the Java changes so the stable cookbooks will stay on 18.1.0 and will file a followup issue to tweak the scripts so we can bump versions separately for Java.

amoeba:
CI looks good, I'll merge in a bit here.

raulcd (approved):

PR #375, opened 2025-02-19
MINOR: [Release] Update versions for 19.0.1

amoeba:


java/source/demo/pom.xml
```diff
@@ -41,7 +41,7 @@
     <properties>
         <maven.compiler.source>11</maven.compiler.source>
         <maven.compiler.target>11</maven.compiler.target>
-        <arrow.version>18.1.0</arrow.version>
+        <arrow.version>19.0.1</arrow.version>
```
lidavidm, line 44: Seems this got messed up?

amoeba: Yeah, I forgot we needed to change the process to separate out Java. I put up a PR for that at https://github.com/apache/arrow-cookbook/pull/376, re-ran the script with the changes in that PR, and force-pushed to this branch.

amoeba:
I redid the branch with the new version of the script in https://github.com/apache/arrow-cookbook/pull/376 so only C++ and Python are updated in this PR.

Issue #351, opened 2024-05-22
[Java] Rework how Java cookbooks are developed and built


amoeba:
In https://github.com/apache/arrow-cookbook/pull/350#issuecomment-2121850653 it was pointed out that the way the Java cookbooks work could be improved quite a bit. We might consider two more recent approaches:

- https://github.com/apache/arrow-adbc/blob/main/docs/source/ext/adbc_cookbook.py
- https://github.com/apache/arrow-adbc/blob/main/docs/source/ext/javadoc_inventory.py

PR #391, opened 2025-11-16
[JAVA][CI] Adds a recipe for Flight SQL

jorisgillis:
A very basic database server is implemented that uses the DataFusion query engine for answering SQL queries on single-file, write-once tables stored in the Arrow IPC format.

Although this does not implement all the features and possibilities of FlightSqlProducer, it show cases the two main categories of concepts through the inclusion of tables (like catalog, database schema, sqlinfo, etc) and statement (like prepared statement and substrait statement).

Some updates to the building of the Java Cookbook: Using the release property instead of the compiler source and target values which is deprecated.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,4 @@ r/*.Rproj
 *target/
 .DS_Store
 .cp.tmp
+python/compressed.csv.gz
--- a/java/source/conf.py
+++ b/java/source/conf.py
@@ -33,11 +33,11 @@ sys.path.append(os.path.abspath("../ext"))
 
 # -- Project information -----------------------------------------------------
 
-project = 'Apache Arrow Java Cookbook'
-copyright = '2022, Apache Software Foundation'
-author = 'The Apache Software Foundation'
-arrow_nightly=os.getenv("ARROW_NIGHTLY")
-if arrow_nightly and arrow_nightly != '0':
+project = "Apache Arrow Java Cookbook"
+copyright = "2022, Apache Software Foundation"
+author = "The Apache Software Foundation"
+arrow_nightly = os.getenv("ARROW_NIGHTLY")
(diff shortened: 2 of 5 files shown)


jorisgillis:
Apparently I did not compile with full Java 11 support (or limitations in this context). I believe the compilation issue is fixed for Java 11.

jorisgillis:
If someone could give me an approval for running, that would be awesome.

raulcd (commented):
This looks reasonable to me but I am neither an Arrow Java nor a Flight SQL expert.
@lidavidm @amoeba do you have some time to take a look at this cookbook recipe?

jorisgillis:
Anything I can do to help move this forward?

lidavidm (commented):
I appreciate the contribution, but I'm wondering if this is really a "recipe" in the first place. In my mind, a recipe is meant to be a short code snippet accomplishing a narrow task, while this is more of a general example project implementing a full-blown database server. 

@raulcd do we have a better place for this kind of example code?

lidavidm:
(I'm also going to hammer on #351 again; having a ton of code embedded into reST makes it hard to author and to review)

jorisgillis:
> I appreciate the contribution, but I'm wondering if this is really a "recipe" in the first place. In my mind, a recipe is meant to be a short code snippet accomplishing a narrow task, while this is more of a general example project implementing a full-blown database server.
> 
> @raulcd do we have a better place for this kind of example code?

I get your point. It is not like reading a CSV where you only have to change the filename. However, I do think it is a recipe, but one with more parameters. 

Nevertheless, I propose to move it to the documentation. I can add more explanation about Flight SQL, build up the example/recipe a bit more in parts. 
What do you think?

jorisgillis:
Closing in favor of [GH-979](https://github.com/apache/arrow-java/pull/980/)

PR #399, opened 2026-04-22
MINOR: [Release][C++][Python] Update versions for 23.0.1

amoeba:
Updates cookbooks from 21.0.0 to 23.0.1.
Bumps the Python version from 3.9 to 3.10 since we dropped support for 3.9 last year.


amoeba:
There was one deprecation to fix in the C++ cookbook and I needed to update the S3 bucket for the Python cookbook now that the Ursa labs bucket is gone. I opted to use the arrow-datasets bucket we use in the main Arrow docs.

